PR #9, opened 2026-07-18
Feature/per output go live

grunt2001c:


zitlem:
## Review

Compiles clean and the UI/state plumbing is solid, but as it stands the core feature isn't wired end-to-end.

### 🔴 Blocking — the selected target never restricts projection
The whole point is "go live to a *specific* output," but the chosen target is captured and never applied. Every reference to the new state is UI-side only:

- `models/GoLiveTarget.kt` — the sealed type
- `viewmodel/PresenterManager.kt` — stores it + `setGoLiveTarget`
- `composables/ProjectionGoLiveButton.kt` — the button reads/writes it for its own display
- `tabs/BibleTab.kt` — reads `.value` to pass into the button

No output renderer (`PresenterScreen`, `DeckLinkComposeOutput`, the screen-assignment path) reads `goLiveTarget`. Each tab still fires the historical all-outputs action:

```kotlin
onGoLive = { presenting(Presenting.STT) }   // ignores the selected target
```

and `TabGoLiveButton.onGoLive` does `setGoLiveTarget(chosen); onGoLive()` — it saves the choice, then projects to everything as before. `containsDisplay()` (the only function that would gate an output) is called only in the dropdown UI to draw the checkmarks.

**Effect:** picking "Display 2" shows "Display 2" on the button but Go Live still lights up *all* outputs. The projection path needs to filter outputs by `goLiveTarget` (with `All` = every output) for this to do what the title says. Is that consumption side a follow-up, or missing here?

### 🟡 No tests
`GoLiveTarget.toggleDisplay` (collapse to All/Display/Multi) and `containsDisplay` are pure, trivially-testable logic with zero tests. Worth adding unit tests for both, especially given the coverage work in flight.

### 🟡 Inconsistent wiring
`BibleTab` wires `ProjectionGoLiveButton` inline (reading `goLiveTarget.value` itself) while every other tab goes through the `TabGoLiveButton` wrapper. Route BibleTab through the wrapper too, or note why it's special.

### ✅ Good
- Clean sealed-type modeling (`All`/`Display`/`Multi`) over a shared index space.
- `toggleDisplay`'s collapse logic reads correctly (0→All, 1→Display, many→Multi).
- New strings added only to the default `values/strings.xml` — no locale files touched.
- The `TabGoLiveButton` wrapper is a nice consistency helper.

**Recommendation:** request changes — solid UI/state groundwork, but it can't merge as "per-output go live" until the projection path honors `goLiveTarget`. Then add tests for `toggleDisplay`/`containsDisplay` and reconcile the BibleTab wiring.

zitlem (changes requested):
Reviewed. The UI is in good shape and the rollout is complete — `TabGoLiveButton` replaces `GoLiveButton` in all twelve tabs, and the 16 new strings are English-only as required. Three things need resolving before this can merge.

**1. The selected target never reaches the output (blocking).**
`GoLiveTarget` is stored and read back, but nothing consumes it when content actually goes live. The only references in the whole branch are `PresenterManager` (the `_goLiveTarget` field plus `setGoLiveTarget`), `ProjectionGoLiveButton.kt` (which reads it to render the trigger label), and `BibleTab`. No presenter, no output-routing path and no screen-lock update reads `goLiveTarget`, so picking "Display 2" and clicking Go Live still projects to every output exactly as before — the button reads as per-output while behaving as All. `GoLiveTarget.Multi` is likewise constructed by the picker and never acted on.

**2. The new button hardcodes a dark palette instead of theming (blocking for light themes).**
`ProjectionGoLiveButton.kt` defines ~25 fixed sRGB colors (`IdleMainBg = Color(0, 47, 61)`, `DropBg = Color(9, 16, 25)`, `TextSelected`, …) and uses them unconditionally. The button it replaces derives from `MaterialTheme.colorScheme.primary`/`onPrimary` (PrimaryActionButtons.kt:104). The app ships nine themes including light ones, so as written the Go Live button and its dropdown render as a dark navy block on a light surface in every tab. These want to come from `colorScheme` (or at minimum branch on `isSystemInDarkTheme()`/the app's theme).

**3. Merge conflict.**
The branch is behind main and conflicts in `tabs/SongsTab.kt`; `viewmodel/PresenterManager.kt` auto-merges. Needs a rebase or merge from main before CI can run — there are currently no check results on this PR at all.

Also worth adding: there are no tests. The picker's pure logic — the toggle that collapses a selection to `All`/`Display`/`Multi`, and the trigger title/subtitle for each of the three cases — is ordinary code that can be tested directly once the target is actually wired to an output.

PR #18, opened 2026-07-24
Render-test BiblePresenter lower-third layouts (presenter/ coverage)

zitlem:
Covers two `BiblePresenter` layout branches the full-screen single-verse tests in `PresenterRenderTest` don't reach (package was ~47%).

## What's covered (2 tests)
- **Lower-third mode** — an entirely separate styling path (every font/colour/size picker switches to its `*LowerThird*` field). Asserts it still composes the verse and its reference; if it stopped, the band would go blank on air while the operator's full-screen preview looked fine.
- **Lower-third + a second Bible** — the secondary-translation branch of that path, which neither the full-screen bilingual test nor the mono lower-third test reaches. Asserts both translations land in the band.

## Notes
- Both assert on verse text / reference, so the layout path runs; nothing races a crossfade.
- Deterministic across repeated `--rerun-tasks`.
- `presenter/` only — no overlap with the dev's open `tabs/` (#9) work (#13 filechooser already merged).


grunt2001c (approved):

grunt2001c:
retest

grunt2001c (approved):

PR #20, opened 2026-07-24
Utils Unit tests

grunt2001c:


zitlem:
Reviewed this — really clean extract-to-testable-seam work across utils/ (the `WindowsRunKey` / `User32`/`GDI32` / `X11` injection and the `Duration.ZERO` retry test are exactly the right patterns). Traced the production refactors and they're all behaviour-preserving.

One issue I went ahead and fixed directly on this branch (commit `c834382a`) — please review and confirm:

**`AutoStartManagerTest` was writing to the real home directory.** The mac/linux paths resolved from the live `System.getProperty("user.home")`, and those are *exactly* the app's production autostart files (`~/Library/LaunchAgents/org.churchpresenter.app.plist`, `~/.config/autostart/churchpresenter.desktop`). So the register/unregister flow **and** the `@Before`/`@After` cleanup wrote to and deleted them:
- A contributor with ChurchPresenter set to launch at login would have had that registration silently removed by a plain `./gradlew check`.
- A JVM killed mid-test (before `@AfterTest`) could leave a bogus plist pointing at a test exe path in their real `~/Library/LaunchAgents/`.

Fix: point `user.home` at a per-test temp dir in `@BeforeTest` (restored in `@AfterTest`), so the production path getters resolve into the temp dir and cleanup only touches the temp tree. `macPlist`/`linuxDesktop` became `lateinit` so they're computed after the swap. All AutoStartManager/LottieFonts/HeicDecoder/ContactReporter/Constants/DevFlags tests pass, verified deterministic 3× with `--rerun-tasks`, and `~/Library/LaunchAgents` / `~/.config/autostart` stay clean.

Non-blocking notes, your call:
- `SongFolderWatcher`'s new `stop()`/`WatchService.close()` is a genuine leak fix but has no test — worth covering or calling out.
- `build.gradle.kts` here adds a jacoco `exclude`; my #21 adds a new task in the same file — different regions, should merge clean, just re-run whoever's second.

LGTM once you've confirmed the isolation change reads right to you.

PR #21, opened 2026-07-24
Test: composables + dialog-overlay render coverage, and a jacoco coverage floor

zitlem:
Continues the coverage push into the View layer.

## Tests (16 new, all <0.3s, verified deterministic 3× with --rerun-tasks)

**composables/** — reusable UI components
- `SearchTextFieldTest` — label + initial query both render
- `SettingsSectionTest` — title header and slot content compose
- `SelectionListTest` — every item renders; clicking reports the exact item
- `ConnectionStatusRowTest` — connected/errored states show the supplied labels

**dialogs/** — inline overlays (no `DialogWindow`, so headless-safe)
- `InstanceLinkToastHostTest` — soft vs hard failure message branch; empty → no toast
- `RemoteActivityToastHostTest` — newest notification's action/title, client attribution, follower badge; empty → nothing
- `CrashFeedbackDialogTest` — Send gated on non-blank comment, sends trimmed values; "Not now" dismisses without sending

These assert branch/wiring invariants (which text wins for which state, callback payloads), not pixel values.

## Coverage floor

Registers `jacocoTestCoverageVerification` with a 75% line minimum (the realistic MVVM target — the View layer is a large, low-logic share of the lines). **Deliberately not wired into `check`** yet: overall coverage is still climbing toward 75% and a gate would block every build until it lands. Run on demand:

    ./gradlew :composeApp:jacocoTestCoverageVerification

A comment in build.gradle.kts documents how to enforce it in CI once the report clears 75%.

## Why dialogs are mostly render-tested indirectly

Most top-level dialogs wrap content in `DialogWindow` (a real AWT window) or enumerate fonts via `GraphicsEnvironment` at composition — both throw headless. This PR targets the dialog composables that render inline. The `DialogWindow`-wrapped bodies are candidates for a later extract-the-content refactor.

### DIFF
--- a/composeApp/build.gradle.kts
+++ b/composeApp/build.gradle.kts
@@ -634,6 +634,41 @@ tasks.register<JacocoReport>("jacocoTestReport") {
     finalizedBy("printCoverageLink")
 }
 
+// Coverage floor — the 75% line-coverage target agreed for this MVVM app. 85% overall isn't a sane
+// goal here: the View layer (tabs/dialogs/composables) is a large share of the lines but low-logic,
+// so chasing the last slices means render-testing near-pure UI wiring. 75% is the honest ceiling.
+//
+// Deliberately NOT wired into `check` yet: overall coverage is still climbing toward 75%, and a
+// failing gate would block every build in the meantime. Run it on demand:
+//   ./gradlew :composeApp:jacocoTestCoverageVerification
+// Once the report clears 75%, enforce it in CI by adding:
+//   tasks.named("check") { dependsOn("jacocoTestCoverageVerification") }
+// Mirrors jacocoTestReport's execution/class/source wiring exactly so the number it gates on is the
+// same one the report prints (app package only; submodules measured by their own builds).
+tasks.register<JacocoCoverageVerification>("jacocoTestCoverageVerification") {
+    group = "verification"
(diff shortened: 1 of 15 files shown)


zitlem:
Added 5 more tests (same branch): render + round-trip coverage for **settings tabs** — `MediaSettingsTab` (both section headers render; toggling Loop drives the real state round-trip and flips only `isLooping`), and full-tab compose smoke tests for `BackgroundSettingsTab`, `AtemSettingsTab`, `LowerThirdSettingsTab` (assert the first section renders from default `AppSettings`). All <1s, verified deterministic 3×. Branch total: **21 tests**.

zitlem:
Added 3 more (branch total **24 tests**): render coverage for the remaining plain settings tabs — `SongSettingsTab` (the largest pure-View surface, ~1.4k lines), `CompanionSatelliteSettingsTab` and `OBSSettingsTab`, each composed from default `AppSettings` and asserting the first control renders. Verified deterministic 3×.

Settings-tab sweep is now complete for the cleanly-testable ones. Not covered (need an extract-content refactor, out of scope here): `SystemSettingsTab`/`StageMonitorSettingsTab` (enumerate fonts via `GraphicsEnvironment` at composition → throw headless) and `ServerSettingsTab` (opens a `DialogWindow`).

grunt2001c (approved):

PR #22, opened 2026-07-24
Refactor: extract BibleTab logic into tested helpers (verse-line, search-highlight, live-nav)

zitlem:
First slice of the fat-tab logic extraction (BibleTab is ~2,150 lines of View, 0% covered).

## What moved
The displayed-verse format — `"16. For God so loved…"` — was re-parsed inline with `substringBefore(". ")` / `substringAfter(". ")` at **~18 sites** across `BibleTab.kt` and `BibleViewModel.kt`, none of it unit-tested. Also duplicated: a verse-range span computation (BibleTab vs the VM's private range parser) and a **triplicated** "first live verse" lookup.

New file `viewmodel/VerseLine.kt` — pure `internal` helpers, every call site routed through them:
- `verseNumberOf(line)` — leading verse number, or null
- `verseTextOf(line, fallback)` — text after `"N. "`
- `indexOfFirstLiveVerse(verses, liveNumbers)` — replaces the triplicated lookup
- `formatVerseReference(line, book, chapter)` — `"John 3:16"` / `"John 3"`
- `verseSpan(range, fallback)` — `(start, end?)` for CCLI/training logging

## Safety
Behaviour is **identical** — mechanical swaps preserving each site's exact semantics (including the empty-string fallback and single-vs-range end handling).
- Full existing `BibleViewModel*` suite: green
- `BiblePresenter*` render tests: green
- New `VerseLineTest` (17 cases, <0.1s, deterministic 3×) covers the invariants those sites relied on

Net: −31 lines at the call sites, and the previously-untested verse-line logic is now behind a name and under test. Groundwork for extracting the heavier BibleTab logic (search-highlight ranges, live-nav math, STT status) in follow-ups.

### DIFF
--- a/composeApp/src/jvmMain/kotlin/org/churchpresenter/app/churchpresenter/tabs/BibleTab.kt
+++ b/composeApp/src/jvmMain/kotlin/org/churchpresenter/app/churchpresenter/tabs/BibleTab.kt
@@ -130,6 +130,7 @@ import org.churchpresenter.app.churchpresenter.viewmodel.DetectionSource
 import org.churchpresenter.app.churchpresenter.viewmodel.ContinuationSpeed
 import org.churchpresenter.app.churchpresenter.viewmodel.formatVerseReference
 import org.churchpresenter.app.churchpresenter.viewmodel.indexOfFirstLiveVerse
+import org.churchpresenter.app.churchpresenter.viewmodel.nextLiveVerseNumber
 import org.churchpresenter.app.churchpresenter.viewmodel.verseNumberOf
 import org.churchpresenter.app.churchpresenter.viewmodel.verseSpan
 import org.churchpresenter.app.churchpresenter.viewmodel.verseTextOf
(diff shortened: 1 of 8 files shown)


zitlem:
Extended this branch with two more pure extractions from BibleTab's composable:
- **`highlightRanges(text, query)`** (utils/TextHighlight.kt) — the case-insensitive match-span finder from the search-result `buildAnnotatedString` loop (empty-query guard + locale-length clamping); the View now only styles the spans it returns. +8 tests.
- **`nextLiveVerseNumber(verses, refVerse, moveUp)`** (VerseLine.kt) — the Up/Down live-panel step math. +6 tests.

Behaviour-preserving; `BibleViewModel*` and `BiblePresenter*` suites stay green. Branch now has **31 new pure test cases** across VerseLineTest (23) + TextHighlightTest (8), all <0.1s and deterministic 3×.

Remaining BibleTab targets for follow-ups: `multiIndicesInFiltered` (best moved onto the VM), and the STT status resolution `when` (many branches).

zitlem:
Added the last BibleTab target — the **STT status ladder**. The inline 10-way `when` over connection booleans is now `bibleSttStatus()` → `BibleSttStatus` enum (viewmodel/BibleSttStatus.kt), with the View mapping enum→stringResource. The ordering (a hard fault outranks a hopeful "listening"; a down engine STT socket is no longer hidden behind it) is covered by `BibleSttStatusTest` (14 cases). Behaviour-preserving; Bible suites green, deterministic 3×.

Branch now has **50 pure test cases** (VerseLine 28 + TextHighlight 8 + BibleSttStatus 14) across all the extracted BibleTab logic. That's every ranked BibleTab target done.

grunt2001c (approved):

grunt2001c:
retest

PR #23, opened 2026-07-24
Refactor: extract SongsTab column-config and display-mode logic into tested helpers

zitlem:
Same fat-tab extraction pattern as #22, now on SongsTab (~2,000 lines of View, 0% covered). SongsViewModel already owns the collection filtering/sorting/section-splitting/navigation — the untested logic here is the table's column configuration and the per-line-mode predicate.

**`utils/SongColumns.kt`**
- `availableSongColumns(songbookCount, hasAddToSchedule)` — which columns exist
- `mergeColumnOrder(saved, available)` — reconcile a saved order against current columns (keep order, drop gone, append new)
- `moveColumn(order, colId, targetId)` — drag reorder; returns same reference on a no-op
- `songColumnSortKey(colId)` — column → sort key
- `draggedColumnIndex(...)` — the off-by-one-prone drag-target math, with column width passed in as a lambda so it's testable without density/Compose

**`utils/SongDisplayMode.kt`**
- `isSongLineMode(settings)` — replaced **three identical** four-way OR-chains across the tab

## Safety
- Full `SongsViewModel*` suite: green
- New `SongColumnsTest` (17) + `SongDisplayModeTest` (5), all <0.1s, deterministic 3×
- No `highlightRanges` duplication — SongsTab does no search highlighting (confirmed)

Independent of #22 (different files); both off `main`.

### DIFF
--- a/composeApp/src/jvmMain/kotlin/org/churchpresenter/app/churchpresenter/utils/SongColumns.kt
+++ b/composeApp/src/jvmMain/kotlin/org/churchpresenter/app/churchpresenter/utils/SongColumns.kt
@@ -1,0 +1,98 @@
+package org.churchpresenter.app.churchpresenter.utils
+
+/**
+ * Table-column configuration for the song list — which columns exist, in what order, how a drag
+ * reorders them, and which sort key each drives. Pure list/string logic extracted from SongsTab so
+ * the table-config decisions can be tested without the Compose table around them.
+ */
+
+/**
+ * The columns available given how many songbooks are loaded and whether the add-to-schedule action is
+ * wired: `"songbook"` only appears with more than one book, `"add_to_schedule"` only when the callback
+ * is present.
+ */
+internal fun availableSongColumns(songbookCount: Int, hasAddToSchedule: Boolean): List<String> = buildList {
+    add("number"); add("title")
+    if (songbookCount > 1) add("songbook")
+    add("tune")
+    add("play_count")
+    add("author")
+    add("composer")
+    if (hasAddToSchedule) add("add_to_schedule")
+    add("favorites")
+}
+
+/**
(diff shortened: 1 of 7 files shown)


composeApp/src/jvmMain/kotlin/org/churchpresenter/app/churchpresenter/utils/SongColumns.kt
```diff
@@ -0,0 +1,98 @@
...
+
+/**
+ * The columns available given how many songbooks are loaded and whether the add-to-schedule action is
+ * wired: `"songbook"` only appears with more than one book, `"add_to_schedule"` only when the callback
+ * is present.
+ */
+internal fun availableSongColumns(songbookCount: Int, hasAddToSchedule: Boolean): List<String> = buildList {
+    add("number"); add("title")
+    if (songbookCount > 1) add("songbook")
+    add("tune")
```
grunt2001c, line 17: id use constants here. real easy to introduce bug with strings

composeApp/src/jvmMain/kotlin/org/churchpresenter/app/churchpresenter/utils/SongColumns.kt
```diff
@@ -0,0 +1,98 @@
...
+    if (fromIdx < 0 || toIdx < 0 || fromIdx == toIdx) return order
+    val mutable = order.toMutableList()
+    mutable.removeAt(fromIdx)
+    mutable.add(toIdx.coerceIn(0, mutable.size), colId)
+    return mutable
+}
+
+/** The sort key a column header drives, or `""` for a column that isn't sortable. */
+internal fun songColumnSortKey(colId: String): String = when (colId) {
+    "number" -> Constants.SORT_NUMBER
```
grunt2001c, line 54: are they the samething?

PR #27, opened 2026-07-24
Refactor: extract ScheduleTab display, zoom and dropped-file logic into tested helpers

zitlem:
Same fat-tab extraction pattern as #22/#23, now on ScheduleTab (~1,300 lines of View, 0% covered). ScheduleViewModel already owns list mutation/undo; the untested logic here is row labelling, the zoom ladder, and dropped-file classification.

**viewmodel/ScheduleItemDisplay.kt**
- `scheduleItemGlyph(item)` — the type→glyph `when` (was **duplicated** at two sites: the row and the drag preview)
- `scheduleItemDetailText(item)` — the grey subtitle, including the 100-char Bible-verse truncation and the `TYPE - path` tags
- `announcementTimerSubtext(item)` — the timer preview, incl. the count-up / live-clock "no fixed preview" rule

**utils/ScheduleZoom.kt** — the card zoom rung ladder: `nearestZoomIndex`, clamped `scheduleZoomIn/Out`, `scheduleCanZoomIn/Out`, and the `scheduleShowCardActions` / `scheduleSingleLineCards` thresholds

**utils/DroppedFileClassifier.kt** — `classifyDroppedFile(ext)` → `DroppedFileAction` (the folder/image-count File I/O stays in the tab; only the pure extension→action decision moves)

## Safety
- Full `ScheduleViewModel*` suite: green
- New: `ScheduleItemDisplayTest` (10), `ScheduleZoomTest` (11), `DroppedFileClassifierTest` (6) — 27 pure cases, all <0.1s, deterministic 3×

## Scope note
ScheduleTab is **not** in PR #9's changed-file set, so this doesn't conflict with the dev's per-output-go-live work. The remaining ScheduleTab target (the drag drop-target/delete-zone hit-test math) is left for a follow-up — it needs the gesture geometry threaded out of `awaitPointerEventScope`, so it's higher-risk than this batch.

### DIFF
--- a/composeApp/src/jvmMain/kotlin/org/churchpresenter/app/churchpresenter/tabs/ScheduleTab.kt
+++ b/composeApp/src/jvmMain/kotlin/org/churchpresenter/app/churchpresenter/tabs/ScheduleTab.kt
@@ -120,7 +120,6 @@ import churchpresenter.composeapp.generated.resources.tooltip_open_schedule
 import churchpresenter.composeapp.generated.resources.tooltip_remove
 import churchpresenter.composeapp.generated.resources.tooltip_remove_from_schedule
 import churchpresenter.composeapp.generated.resources.tooltip_save_schedule
-import kotlin.math.abs
 import kotlinx.coroutines.launch
 import org.churchpresenter.app.churchpresenter.composables.TooltipIconButton
 import org.churchpresenter.app.churchpresenter.data.settings.PlanningCenterSettings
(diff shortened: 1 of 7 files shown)


composeApp/src/jvmMain/kotlin/org/churchpresenter/app/churchpresenter/tabs/ScheduleTab.kt
```diff
@@ -420,21 +422,20 @@ fun ScheduleTab(
...
+                    enabled = scheduleCanZoomOut(itemZoomPercent),
                     buttonSize = 32.dp,
                     iconTint = MaterialTheme.colorScheme.onSurface
                 )
                 TooltipIconButton(
                     painter = painterResource(Res.drawable.ic_zoom_in),
                     text = stringResource(Res.string.tooltip_schedule_zoom_in),
-                    onClick = { onItemZoomChange(ZOOM_LEVELS[zoomIndex + 1]) },
-                    enabled = zoomIndex < ZOOM_LEVELS.lastIndex,
+                    onClick = { onItemZoomChange(scheduleZoomIn(itemZoomPercent)) },
```
grunt2001c, line 436: should not need to make modifications to compose code if you use compose ui tests and test tags attached to modifiers you should be able to reach most of the ui elements and test them out.

grunt2001c (changes requested):

PR #34, opened 2026-07-25
Refactor: extract + test v4l2 camera-format parsing

zitlem:
`listV4l2Formats` (in SourcePropertiesPanel) shells out to `ffmpeg -list_formats` and `v4l2-ctl --list-formats-ext` and parsed their text output inline with regexes — pure, error-prone logic that had no tests. Independent of every open PR (#9 doesn't touch this file).

The process calls stay in `listV4l2Formats`; only the parsing moves out:
- `parseFfmpegV4l2Formats(output)` — each `WxH` line, with a same-line `N fps` or a 30 default.
- `parseV4l2CtlFormats(output)` — `WxH` on its own line, fps on the following lines (each fps pairs with the most recent size).
- `sortedCameraFormats(sizes)` — largest area first, then highest fps; mapped to the display model.

`CameraFormat` widened `private → internal` (no behaviour change).

## Tests — `CameraFormatParsingTest` (6)
Representative tool output for each parser (size-with/without-fps, non-resolution lines skipped, empty output; the size-then-fps pairing and the ignore-fps-before-any-size rule), plus the area-then-fps ordering. Deterministic 3×.

### DIFF
--- a/composeApp/src/jvmMain/kotlin/org/churchpresenter/app/churchpresenter/composables/SourcePropertiesPanel.kt
+++ b/composeApp/src/jvmMain/kotlin/org/churchpresenter/app/churchpresenter/composables/SourcePropertiesPanel.kt
@@ -1140,13 +1140,57 @@ private fun listCameraDevicesWithDeckLink(deckLinkDeviceFormat: String = "DeckLi
     return devices
 }
 
-private data class CameraFormat(
+internal data class CameraFormat(
     val width: Int,
     val height: Int,
     val fps: Int,
     val displayName: String = "${width}x${height} @ ${fps}fps",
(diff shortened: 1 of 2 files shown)


zitlem:
Superseded by #47, which refactors and tests the same `SourcePropertiesPanel` camera-format parsing — plus the dshow, avfoundation, Windows/Mac device-listing and window-listing paths this PR left alone. Its `SourcePropertiesDeviceListingTest` covers everything `CameraFormatParsingTest` did (the size-with/without-fps cases, the v4l2-ctl size-then-fps pairing, the fps-before-any-size rule and the area-then-fps ordering) and more, so nothing here is lost. The two touch the same functions and conflict, so closing this one rather than asking #47 to rebase.

PR #37, opened 2026-07-25
Background Settings Unit tests

grunt2001c:


zitlem:
Nice, thorough Background coverage. The CI red is one root cause — **no VLC on the CI runner**, not a flake.

`BackgroundTypeDropdown` gates the Video option on `isVlcAvailable`: when VLC is absent it renders the item **disabled** and relabels it `Video Loop (Install VLC)` (`disabledTypes = if (!isVlcAvailable) setOf(BACKGROUND_VIDEO)`; the menu item's `enabled = !isDisabled`). CI runners have no VLC, so the three failing tests all try to reach the Video type and can't:

- `choosing the video background type reveals the video picker` and `the Songs full-screen column stores the type it is given` → `chooseBackgroundType(..., TypeLabel.VIDEO)` clicks a **disabled** menu item (whose label is actually `Video Loop (Install VLC)`), so nothing is selected and the `BACKGROUND_VIDEO` assertion fails.
- `each dropdown menu offers exactly the types its slot supports` → the exact-text/count for `Video Loop` shifts to `Video Loop (Install VLC)`.

You already handle this in the labels test (the `if (isVlcAvailable) … else "(Install VLC)"` branch, ~line 1358) — the three failures are the spots that don't. Options: guard those on `isVlcAvailable`, or set the Video type via the initial `AppSettings` (bypassing the disabled dropdown), or make `isVlcAvailable` injectable so tests can force it. Reproduce locally by ensuring VLC isn't on the path.

One heads-up: this PR rewrites `BackgroundSettingsTabTest.kt` and `SongSettingsTabTest.kt`, which my open #30 also edits — we'll collide. Yours is far more thorough on Background, so I'll slim #30 down to just its non-Background/Song tabs to get out of your way (see the PR thread).

PR #45, opened 2026-07-26
Companion Sattelite unit tests

grunt2001c:


zitlem (approved):
Approve. CI green, headless-safe (no VLC/socket at construction), and additive (new split test files) so no conflict with my open #30. Thorough Companion Satellite coverage — connections, fields, placement, recomposition, and the view-model-gated half. Two non-blocking notes: it overlaps my #30's CompanionSatelliteSettingsTabTest (yours is more thorough — I'll slim #30 to drop it), and UploadBackgroundToAtemTest is also added by #37, so whichever of those merges second will conflict on that one file.

PR #61, opened 2026-07-27
Detect which Bible translation is being read aloud

zitlem:
Adds Bible **version** detection to the speech-following feature: which *translation* the speaker is reading from, shown on each detection row.

The engine side lives in ChurchPresenter-BLE and is already on `master` there (`282f6b8`). This branch is the ChurchPresenter half and **compiles against the currently pinned BLE commit** — the feature simply stays quiet until the pointer bump lands, so the two can merge in either order. The bump is a separate PR per the submodule rules.

### What it does

Scores the spoken words against every `.spb` in the Bible folder, not just the two loaded for detection — the reader's version is frequently one the operator hasn't got selected. Against the real 19-version corpus here it names NASB from the second verse of Matthew 18 and holds it for the rest of the service.

The answer arrives on its own `version_detected` message rather than on a scripture event. The engine needs a verse or two of reading before it can tell, so it routinely lands *after* the rows it describes are already on screen; those rows are backfilled, while rows the engine stamped itself are left alone.

### Also here

**Bibles in subfolders are found.** Every listing used `listFiles` on the root alone, so pointing the app at a real nested collection found only what sat at the top level — 3 of 71 modules in the folder used for testing. The listing now walks recursively and returns paths relative to the storage directory; previously stored bare filenames keep resolving unchanged. This also fixes a documented gap where a *folder* named `x.spb` was offered in the picker as a module.

**The Help Dev flag pills work again.** They were drawn in `surfaceVariant` with icon and label at `alpha 0.6` — this app's idiom for a disabled control — with no press feedback and no visible effect, since a click only appends a line to a log. The flag log shows an operator clicking the same one seven times in 1.4 seconds. They now carry real colour, flash on click, and *Wrong passage* / *Premature* go properly disabled when nothing is live, which is exactly when they used to swallow the click silently.

**`onScripture` is now a data class.** It was an 11-argument positional lambda with four adjacent `String?` parameters the compiler cannot distinguish, and this change adds more — a transposition would compile cleanly and silently corrupt the training-log join keys.

### Verification

`./gradlew :composeApp:check` — 4783 tests, 0 failures. New coverage for nested listing, title disambiguation, `version_detected` parsing, and the row backfill.

Unrelated pre-existing flake seen once during this work: `PlanningCenterDownloadTest :: a thumbnail from a server error comes back as nothing` (passes 3/3 in isolation).


zitlem:
Reviewed — merging.

The `FileManager` recursive listing is the part I looked at hardest, since it changes the identity everything keys on. It holds up: bounded depth stops a symlink cycle hanging the settings dialog, `it.isFile` closes the documented gap where a *folder* named `x.spb` was offered as a module, and `toRelativeString` of a file at the root is still just its name — so a stored bare `primaryBible` keeps resolving and there is no migration step.

`onScripture` → `EngineScripture` is well motivated. Eleven positional arguments with four adjacent `String?` in the middle is a transposition waiting to happen, and the ones at risk (`sessionId`, `segmentId`) are exactly the join keys where a silent swap would corrupt the training log rather than crash.

The backfill is correct, including the part that is easy to get wrong: it early-returns when no row needs filling, so an arriving `version_detected` that changes nothing does not emit new state and trigger a pointless recomposition of the list.

## One defect

There are two adjacent KDoc blocks before the class:

```kotlin
 * ... The level chip is pushed to the engine via [setLevel].
 */
/**
 * @param onVersion the translation the engine believes is being read aloud, ...
 */
class BibleEngineClient(
```

Only the second one attaches to `BibleEngineClient`. The first — the actual description of what the class *is* — becomes an orphaned comment, so it will not show in KDoc output or in an IDE tooltip. It looks like the new `@param` block was inserted ahead of the class rather than merged into the existing block.

The fix is to delete the intervening `*/` + `/**` so it is one block again. Not merging past it because it is broken, only because it is non-functional and this branch is checked out in a working tree with in-progress changes — pushing a commit to it from outside would be more disruptive than the defect. Worth folding into the next push on this file.

Issue #24, opened 2026-07-24
Flaky: disk-loading SongsViewModel tests time out under JVM contention


zitlem:
### Summary
The disk-backed `SongsViewModel` unit tests fail intermittently (~50%) when run together in one JVM, and pass when run alone. The failure is always:

```
java.lang.AssertionError: timed out after 5000ms waiting for songs
```

i.e. the test's `awaitUntil { vm.filteredSongItems.value.size == 3 }` never sees the songs load within 5s.

### Affected tests
All the ones that build a real `SongsViewModel` against a temp song directory:
- `SongsViewModelSortingTest`
- `SongsViewModelLibraryTest`
- `SongsViewModelNavigationEdgeTest`

(`SongsFilteringTest` is **immune** — it injects songs via `setInstanceLinkSource`, so it never hits the async disk load.)

### Reproduce
```
./gradlew :composeApp:jvmTest --tests 'org.churchpresenter.app.churchpresenter.viewmodel.Songs*' --rerun-tasks
```
Run a few times — it fails on some runs, passes on others. Different tests fail each time; the common factor is the 5s load timeout, not any specific assertion.

### Root cause (as far as I got)
Each of these tests constructs a real `SongsViewModel`, which kicks off an async `loadSongs()` on `viewModelScope` **and** starts a real `SongFolderWatcher` (a coroutine blocking an IO thread). With several such view models loading concurrently across the suite, the loads contend on the dispatcher / IO threads and the 5s `awaitUntil` window is missed. It's contention + a tight timeout, not a wrong assertion.

### Ruled out
I suspected the song cache — it's a **single shared file** at `~/.churchpresenter/song_cache.json` (`SongFileParser` companion, a latched `val`, not per-storage-dir), and these tests don't isolate `user.home`. I tested that hypothesis (made the path a live `get()`, isolated `user.home` in `SortingTest`) and it **did not** fix the timeout — so the shared cache is not the cause.

### Suggested fix (needs an owner familiar with the async load)
A test seam on `SongsViewModel` to either:
- load songs synchronously in tests, or
- construct the VM without the `SongFolderWatcher`,

so the tests don't race a real async load + watcher. Bounding/awaiting the load deterministically (a positive "loaded" signal that also survives contention) would remove the flake without widening the timeout. Flagging rather than patching since this is the async-load design.

zitlem:
This is now hitting real PR CI, not just local runs. On PR #30 the `test` check failed with:

```
[...]

PR #67, opened 2026-07-27
Make the SongsViewModel tests synchronous, and fix the stale-cache bug it exposed

### LINKED ISSUE #56: Three SongsViewModel file-watcher tests are flaky — they fail by timeout expiry
Confirmed flake, with evidence: run `30224946031` on PR #53 went **red then green on the identical commit**. The only difference was a re-run of the same job.

## The failures

```
SongsViewModelEditDeleteTest > a lyrics-only edit keeps the very same file
    java.lang.AssertionError: timed out after 5000ms waiting for reload
    at SongsViewModelEditDeleteTest.awaitUntil(SongsViewModelEditDeleteTest.kt:68)

SongsViewModelLibraryTest > renaming a song renames its file
    java.lang.AssertionError: timed out after 5000ms waiting for the rename
    at SongsViewModelLibraryTest.awaitUntil(SongsViewModelLibraryTest.kt:81)

SongsViewModelLibraryTest > changing the number renames the file too
    java.lang.AssertionError: timed out after 5000ms waiting for the renumber
    at SongsViewModelLibraryTest.awaitUntil(SongsViewModelLibraryTest.kt:81)
```

All three failed together, all by the 5 s deadline expiring rather than by an assertion — the signature of a loaded runner outrunning a filesystem watcher, not of a defect.

## Why it is worth fixing rather than re-running

`AGENT.md` calls this out directly:

> A wait must end on a **positive signal** — the state you expect, a frame you can identify, a callback firing. It must never end by the timeout expiring; the timeout exists only to fail the test. A `waitFor { ... } == false` result costing the full timeout is the bug this rule is about.

and:

> **No flaky tests.** A test that passes only most of the time is worse than no test: it trains everyone to re-run instead of to read the failure.

Each failure also costs the full 5 s, so a bad run burns 15 s of CI before failing.

## What to look at

`awaitUntil` in both classes polls to a deadline. These three wait on `SongFolderWatcher` picking up a change on disk. The fix is to end the wait on a signal the watcher itself emits — a callback or an observable state transition — rather than polling until 5 s runs out. Where the watcher genuinely cannot signal, `AGENT.md`'s guidance is to drop the test and record the gap in the class doc comment instead.

## Not urgent

Nothing is broken in the product; this is test reliability. Recording it so the next red build on these three is read as "the known flake" and fixed, rather than silently re-run.

zitlem:
Closes #56.

## The flake

These tests polled a wall clock for up to 5s waiting for a load. The view model's `dispatcher` parameter only backed `viewModelScope` — the work itself ran on a hardcoded `withContext(Dispatchers.IO)`, so **no test-supplied dispatcher could reach the part that does the work**. On a loaded runner the coroutine simply had not been scheduled inside 5s and the deadline expired. Red-then-green on one commit, three times in a session.

## The fix

`ioDispatcher` becomes a constructor parameter alongside `dispatcher`, and the tests pass immediate dispatchers for both. A load then completes before the constructor returns, so there is nothing to wait for — the polling helpers and one hand-rolled `Thread.sleep` loop are replaced by a single assertion that the work *already happened*.

I verified that claim before committing to it, by making the helper check exactly once and confirming all 66 still pass. **66 tests, 2.1s → 0.32s, and no clock anywhere in them.**

## The bug this exposed

Making it synchronous turned an intermittent product bug into a deterministic failure:

```kotlin
if (cached != null && cached.lastModified == lastMod) {
    results.add(cached)   // reuse
}
```

`File.lastModified()` resolves to milliseconds at best, and a **full second** on some filesystems. A song edited within the same tick as its cache entry looks unchanged, so the edit is dropped on the next load — the user's second quick edit silently vanishing until something else invalidates the cache. The async IO path used to take long enough to usually cross a tick, which is why this was never seen.

`CachedSong` now carries `fileSize` and both must match.

**Limits, stated plainly:** this catches any edit that changes the file length, which is nearly all of them. A *same-size, same-tick* edit would still slip through and would need a content hash to catch — I did not do that, because hashing every song on every load is a real cost for a much narrower case. `fileSize` defaults to `0` so caches written by an older build still deserialize; those entries miss once and get re-parsed.

Full `:composeApp:jvmTest` green.

### DIFF
--- a/composeApp/src/jvmMain/kotlin/org/churchpresenter/app/churchpresenter/data/SongFileParser.kt
+++ b/composeApp/src/jvmMain/kotlin/org/churchpresenter/app/churchpresenter/data/SongFileParser.kt
@@ -12,7 +12,17 @@ import java.nio.file.Paths
 @Serializable
 data class CachedSong(
     val song: SongItem,
-    val lastModified: Long = 0L
+    val lastModified: Long = 0L,
+    /**
+     * File size, checked alongside [lastModified] to decide whether a cached parse is still good.
+     *
+     * Timestamp alone is not enough: `File.lastModified()` has millisecond resolution at best and a
+     * whole second on some filesystems, so a song edited within the same tick as the cache entry
+     * looks unchanged and the edit is silently dropped on the next load. Size catches that for any
+     * edit that changes the length, which is nearly all of them. Defaults to 0 so caches written by
+     * an older build still deserialize — those entries simply miss once and get re-parsed.
+     */
+    val fileSize: Long = 0L,
 )
 
 @Serializable
@@ -231,16 +241,17 @@ class SongFileParser {
         for (songFile in songFiles.sortedBy { it.name }) {
             val path = songFile.absolutePath
             val lastMod = songFile.lastModified()
+            val size = songFile.length()
             val cached = cache[path]
 
-            if (cached != null && cached.lastModified == lastMod) {
+            if (cached != null && cached.lastModified == lastMod && cached.fileSize == size) {
                 // File unchanged — reuse cached parse result
                 results.add(cached)
             } else {
                 // File new or modified — parse it
                 val song = parseSongFile(path, songbook)
                 if (song != null) {
-                    results.add(CachedSong(song = song, lastModified = lastMod))
+                    results.add(CachedSong(song = song, lastModified = lastMod, fileSize = size))
                 }
             }
         }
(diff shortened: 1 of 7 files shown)


zitlem:
Worth recording: **this flake was reported and closed before**, as #24 (closed 25 July) — same tests, same `timed out after 5000ms waiting for songs`, same diagnosis of dispatcher contention. It came back as #56.

#24 asked for one of two things:

> A test seam on `SongsViewModel` to either:
> - **load songs synchronously in tests**, or
> - construct the VM without the `SongFolderWatcher`

Only the second was done. `enableFolderWatcher = false` landed, along with `dispatcher = Dispatchers.Default` — which looks like the first one but is not, because the load's actual work sits behind a **hardcoded** `withContext(Dispatchers.IO)`. The seam reached the scope but never the work, so the tests still waited on a shared pool and still polled a wall clock for it. That is why it recurred.

This PR does the half that was missing: `ioDispatcher` is injectable, so a load genuinely completes synchronously and there is nothing left to wait for. The polling is deleted rather than tightened.

One correction to #24 while I am here — it says:

> I suspected the song cache … I tested that hypothesis … and it **did not** fix the timeout — so the shared cache is not the cause.

That conclusion was right about the *timeout*, and the cache is indeed not the contention cause. But the cache did turn out to have a real defect of its own, which only became visible once the load was synchronous: an entry was reused whenever `cached.lastModified == file.lastModified()`, so an edit landing in the same filesystem tick was silently dropped. Two separate bugs sharing one neighbourhood, which is presumably why the earlier investigation kept circling it.

PR #72, opened 2026-07-27
Download Bible translations in the app

grunt2001c:
Getting a Bible into ChurchPresenter meant finding a `.spb` yourself. The app
shipped one bundled KJV and a button to copy it; the setup wizard told people to
"search the internet for a Bible XML file". This adds a browser that downloads a
translation and converts it on the user's machine.

**Requires ChurchPresenter/ChurchPresenter-Converter#1 — merge that first.**
The gitlink here points at `feat/zefania-import`, and CI checks out submodules,
so this branch cannot go green until that PR lands and the pointer moves to main.

## What it does

Settings → System → Bible Storage Directory → **Download Bibles…**

Two tabs, one per archive:

- **eBible.org** — 1,291 translations across 1,023 languages
- **Zefania archive** — 264 across 63

Each tab keeps its own search and scroll position and loads only when first
opened. Picking a translation downloads it, unpacks it, converts it and moves it
into the Bible folder, with the phase and progress on the row; it then appears in
the Primary/Secondary dropdowns immediately.

## Licensing

Every download is confirmed, and the two archives are described differently
because they differ:

- eBible publishes a `Redistributable` flag and a copyright string per entry.
  **Only entries marked redistributable are listed**, and the copyright shows in
  the row before anything is downloaded.
- Zefania publishes neither. Its note says the archive states no licence details
  and its modules are unchecked — rather than claiming the translation has no
  licence, which is a different and misleading assertion.

There is deliberately no "don't ask again": the licence differs per translation,
so consent to one says nothing about the next.

## Verified against the live archives

Not fixtures alone. Both catalogues were fetched and the Berean Standard Bible
installed end to end — all four phases, 66 books, reloading correctly in the
app's own parser with no scratch files left behind.

The USFX parser was the risk, so it was reconciled verse by verse: **all 31,086
verses of the BSB match the source exactly**, and Russian, Arabic (RTL), Greek
and Chinese modules reconcile too. That reconciliation found a real bug — the BSB
wraps Zechariah 12:1 in a descriptor element the parser was skipping wholesale,
silently losing the verse.

Filename parsing was also checked against all 264 real Zefania names: the first
rule missed 16 of them (underscore dates, whole-word language tokens, non-ASCII
replaced by underscores). Czech was installing as `CZE_EKLAD`. Now zero misses,
and those shapes are a test case.

## Structure

`BibleSource` is a two-method interface (`catalog`, `install`) over a shared
`BibleModule`, so the view model is source-agnostic and its tests use a plain
fake rather than mocking a singleton. Download, verification, unzip and the
atomic move are shared in `BibleInstallSupport`.

Integrity is end to end: Zefania downloads are checked against the git blob hash
that comes free with the tree listing, which also rejects the two bodies a file
host returns instead of a module — a 404 page and an LFS pointer. Nothing
touches an installed Bible until a verified replacement exists, so re-downloading
the translation currently on screen cannot destroy it if the connection drops.

Catalogues are cached for a week on disk with ETag revalidation, and a failed
fetch falls back to the last one seen, flagged stale — a hall with no internet
still gets a list. GitHub allows 60 unauthenticated requests an hour per address,
which a church shares.

## Also

- "Add Bible Samples (KJV)" is gone — the browser offers nine King James editions
  among 264. First-run bundling of the KJV stays, so the app still works offline
  out of the box.
- The wizard's Bible step described a `.db3` format the app has not read in a
  long time and pointed at the wrong tab.
- String resources: backslash escapes on apostrophes and quotes were rendering
  literally, backslash and all.

## Tests

94 new tests across six classes; `:composeApp:check` green; three consecutive
reruns clean; slowest new test 0.3s.

Not covered: the dialog composable itself, matching the existing boundary
(`StockMediaBrowserDialog` has no test either).


zitlem:
Reviewed. The feature itself is carefully built — two findings worth fixing before merge, and some things worth stating explicitly because they are done *right* and a reviewer might assume otherwise.

## 1. `BibleInstallSupport.kt` is a binary file to git — nobody can review it

It contains two literal `0x00` bytes, because the git-blob hash is written with a raw NUL rather than an escape:

```
offset 5663:  digest.update("blob ${file.length()}[NUL]".toByteArray(…
offset 5469:  the same, inside the doc comment above it
```

Consequences, all silent:

- **The PR shows `0+ 0-` for this file.** Git treats it as binary, so there is no diff — 6.8 KB of download / extract / atomic-move logic went through review unreadable. I had to `awk` it out to review it at all.
- `grep` skips it by default. A codebase-wide search — including `cleanup_check.sh`'s own greps — does not see this file.
- No blame, and no line-level merge if two branches ever touch it.

The fix is one escape each: `"blob ${file.length()}"`, and write `\0` textually in the comment. Behaviour is identical — the same string is produced.

(Fittingly, posting this review failed the first time because quoting the raw byte tripped a control-character guard.)

## 2. The flake pattern from #24 / #56 is being reintroduced

`BibleCatalogViewModelTest` polls a wall clock:

```kotlin
private fun awaitUntil(what: String, timeoutMs: Long = 5_000, condition: () -> Boolean) {
    val deadline = System.currentTimeMillis() + timeoutMs
    while (System.currentTimeMillis() < deadline) {
        if (condition()) return
        Thread.sleep(5)
```

and `BibleCatalogViewModel` hardcodes its dispatcher:

```kotlin
private val viewModelScope = CoroutineScope(Dispatchers.Main + SupervisorJob())
```

That is the same shape `SongsViewModel` had before #67, and the same failure follows. `Dispatchers.Main` on Compose Desktop is the single Swing event thread, so with the suite loaded the coroutine is not scheduled inside 5 s and the deadline expires.

This exact thing was reported as #24, "fixed" by adding `enableFolderWatcher` plus a scope dispatcher, **came back as #56** because the actual work was still behind a hardcoded dispatcher, and was only removed in #67 by making the inner dispatcher injectable — at which point the load completed synchronously and the polling could be deleted outright rather than tuned.

Small change, and it stops a third round:

```kotlin
class BibleCatalogViewModel(
    private val source: BibleSource,
    private val storageDirectory: String,
    dispatcher: CoroutineDispatcher = Dispatchers.Main,
) {
    private val viewModelScope = CoroutineScope(dispatcher + SupervisorJob())
```

Tests then pass an immediate dispatcher and assert directly — no `awaitUntil`, no clock. `AGENT.md` is explicit that a wait must end on a positive signal and never by timeout expiry.

## What is done right — checked, not assumed

- **Zip-slip is structurally impossible** rather than merely guarded: entries are written to a generated `File(into, "entry$index")` and the archive's own path is never used as a destination. Plus `MAX_ARCHIVE_ENTRIES = 64` and `MAX_EXTRACTED_BYTES = 256 MB` against zip bombs.
- **The redistributable filter is a hard exclusion**, not a display flag — non-redistributable rows are dropped in `mapNotNull` before they can ever be listed.
- **And it fails closed.** If eBible renames that column, `index()` returns −1, `getOrNull(-1)` is null, `isTrue()` is false, and the catalogue comes back **empty** rather than unfiltered. That is the right direction for a licensing gate, and worth a test of its own so it stays that way.
- Tests reach no real network — the eBible and GitHub URLs are production constants with `?:` overrides.
- The scratch directory living inside the Bible folder rather than the system temp dir, so the finishing move is a same-filesystem atomic rename, is a genuinely good call for folders on network shares or external drives.

## On the stated dependency

The description says this "cannot go green until that PR lands". It is green now — I pushed `feat/zefania-import` to the Converter remote, so CI can fetch the pinned SHA.

**Green CI does not mean the dependency is resolved.** Main would still pin a Converter *feature branch*, which dangles if that branch is later deleted or force-pushed. The two Converter commits still need to reach the Converter's `main` before this merges.

zitlem:
Both findings fixed in 5de17751.

**1. The NUL bytes are gone.** The git-blob header now uses the Kotlin unicode escape rather than a raw 0x00, with a comment saying why it must stay one. Same bytes on the wire, same hash. `file` now reports the source as UTF-8 text, and `grep` can see it again.

Worth noting the transition itself still shows as a binary diff in this PR — git compares against the old blob, which *was* binary. From the next commit onward it diffs normally.

I also added **BibleInstallSupportTest**, because nothing tested `gitBlobSha1` at all — and that function is what verifies a Zefania download against GitHub's tree listing, so a wrong hash rejects every download silently. The expected values come from `git hash-object` on the same bytes, not from running our implementation, so it checks against git rather than recording our own output: ordinary content, content with no trailing newline (different byte length, different blob), and the well-known empty-blob hash `e69de29…` which proves the header is written even with nothing after it.

**2. The dispatcher is injected.** `BibleCatalogViewModel` now takes `dispatcher: CoroutineDispatcher = Dispatchers.Main`, and the tests pass an immediate one.

Every assertion then holds synchronously, so I deleted the polling rather than shortening it — and verified that claim before committing to it, by making the helper check exactly once and confirming all 17 still pass. **5s deadlines to 0.079s for the class.**

One thing I checked rather than assumed: `a load while one is already running is not duplicated` waits for `isLoading` to become *true*, which looks like it could not survive a synchronous dispatcher. It does — the test parks the source on a `CompletableDeferred`, so `Unconfined` runs inline up to that suspension and `isLoading` genuinely is true there; completing the deferred resumes inline to false. That test's signal was already the work itself rather than elapsed time, which is why it needed no restructuring.

Full `:composeApp:jvmTest` green, deterministic 3x.

PR #74, opened 2026-07-27
Test the Bible tab

zitlem:
`BibleTab` was 0%-covered because it needs a real `BibleViewModel`, whose scope and file reads were hardcoded to `Dispatchers.Main`/`Dispatchers.IO` — a test could only poll a wall clock for the load, the shape that made the Songs tests flaky (#56). This gives `BibleViewModel` the same two dispatcher seams `SongsViewModel` already has, so the module loads synchronously and the tab is ordinary Compose to drive. Production defaults are unchanged.

## Tests

31 tests over the real load path — a real `.spb` fixture written to disk and read back through the real parser, nothing stubbed, no host window / `PresenterManager` / STT needed:

- **Browse** — the three columns and what each click changes, the column headers, the no-primary-Bible empty state, and the swap button (present only with a secondary Bible, and the settings it produces).
- **Search** — the smart box in all three modes: a reference navigates, free text searches, Text mode forces a search of a reference-shaped query; the result list, clicking a result, no-results, and clear.
- **Go Live / Add to Schedule** — what the host is handed, plus the history panel: ordering, dedup, navigating back from an entry, and clearing it.
- **Split browse** — the live chapter panel mirrors the live chapter while the browser moves away, and clicking a verse in it goes live.

Each test drives the control and asserts the effect (the verses on screen, what the host was handed) rather than that a callback fired. Whole class runs in ~1s, slowest test 0.17s; passes 3× consecutively with `--rerun-tasks`.

## Also here

- **Bugfix (first commit):** every Bible search result printed its reference twice — "Psalms 23:1 - Psalms 23:1 The LORD is my shepherd…". `Bible.addSearchResult` deliberately builds `verseText` with the reference prefix (an existing test pins that), and `BibleTab` prefixed it again when rendering the row.
- One `testTag` on the search field's clear button — its icon is decorative, so there is no label to address it by.

## Not covered

Ctrl/Shift multi-verse selection and the right-click context menu (they read keyboard modifiers and the secondary button inside `awaitPointerEventScope`), and the STT/auto-follow detection panel, which needs a live `STTManager`/`BibleEngineClient`.


grunt2001c (approved):

zitlem:
Reviewed — looks good, no changes requested. (GitHub blocks approving one's own PR, so this comment stands in for the approval.)

- The search-row change is a genuine bug fix, not just a test accommodation: `Bible.addSearchResult` already builds `verseText` as `"$bookName $chapter:$verse ${bv.verseText}"` (Bible.kt:478), so the tab's `"${result.book} ${result.chapter}:${result.verse} - "` prefix printed the reference twice on every result row.
- The `dispatcher`/`ioDispatcher` seams are constructor defaults that reproduce the shipping behaviour and mirror the pair `SongsViewModel` already takes for issue #56 — no mutable singleton seam, and they remove the wall-clock polling that made the Songs tests flaky.
- Tests build real `.spb` modules through `SpbFixture` and read them back through the real load path, and assert on what the tab reported (selected verses, schedule adds, settings transforms) rather than on stub invocations. One `testTag` added, on a decorative icon button with no label to address it by.

CI green. Branch is behind main but merges cleanly.

PR #76, opened 2026-07-27
Test the companion server's lower-third, ATEM-key and live-state paths

zitlem:
Two areas of `CompanionServer` that no existing test reached. Same harness as the sibling suites — a real server on a free port, driven over real HTTP/WebSocket — because `start()` builds its own Netty server rather than exposing a separable Ktor module.

## Lower thirds and the ATEM key (16 tests)

The routes a Stream Deck presses. These reach hardware, so what is pinned is everything the handlers decide **before** that happens: which file a preset name resolves to (case-insensitively, since buttons are configured by hand), which files count as Lotties and what each one's duration is, and whether the requested key target exists on the detected switcher topology. Each of those returns its own status and stops, so no test opens a connection to an ATEM or starts a key sequence.

A wrong answer at this stage is a button that silently does nothing mid-service, or a key cut on a bus that isn't there.

## Live state (8 tests)

`updateLiveState` and the `live_state_changed` frame — the one message a phone's "now showing" view and an Instance Link follower's mirrored output are both built from. Covers the connect snapshot a client joining mid-service receives, the field mapping for Bible, song, picture and website content, and the de-duplication that stops a redraw storm flooding the shared broadcast buffer and evicting messages for slow clients.

## Not covered, on purpose

Two paths whose cost is a wait rather than work, both noted in the test class doc comment:

- A *successful* lower-third run/show hands off to `LowerThirdSequencer`, which talks to a switcher and holds the animation for its full duration. `LowerThirdSequencerTest` covers the sequencer itself.
- The undetected-topology key path (key counts of 0 — the app has never connected) is accepted by validation by design and so has to reach the switcher. With no ATEM on the network that is a 5-second socket timeout; I wrote it, measured it, and removed it.

Both suites run in about a second each. Full suite passes 3× consecutively with `--rerun-tasks`.


grunt2001c (approved):

zitlem:

- Test-only, driven against a real `CompanionServer` over real HTTP/WebSocket, consistent with the existing `CompanionServer*Test` suites. Port 39_715/39_713 don't collide with the ones already in use (39_517, 39_600, 39_620, 39_640, 39_660, 39_711).
- `snapshotLiveState` stops on the live-state frame rather than on the idle window, so the normal path costs the frame arriving; the 250 ms window only has to cover "there is no live state at all", which is within what AGENT.md allows for a snapshot with no final frame.
- Assertions are on the served payload, not on stub invocations.

Nit, non-blocking: `kotlinx.coroutines.delay` is called fully qualified in the startup poll — worth importing it for consistency with the rest of the file.

CI green.

PR #77, opened 2026-07-27
Test how a schedule crosses the wire in both directions

zitlem:
Follow-up to #76, same area, independent branch off main.

`updateSchedule` flattens twelve `ScheduleItem` types into one wide DTO, and a phone builds its schedule list from nothing else — so a mis-mapped field is silent rather than a compile error, and surfaces as an item that is blank or unlabelled on every connected device. `parseRemoteItem` is the way back: it accepts three shapes for historical reasons (the flat companion format, the same format carrying only an id for content the server already knows about, and the legacy sealed-class format), and anything it cannot parse must be refused rather than added as something else.

## Tests (17)

**Outbound** — the DTO mapping per item type: song, bible (a single verse sends no range, a passage does), label with its colours, media, lower third, announcement, countdown (every field a phone needs to run the clock itself), website, scene, dictionary; plus ordering and clearing. Pushing a picture folder is also what registers it, so that test asserts the folder becomes browsable by index afterwards — waiting on the image actually being fetchable rather than on a timer, since registration happens off the calling thread.

**Inbound** — a song and a verse in the companion format, a picture sent as just a `folder-id` resolved against the catalogue, the legacy discriminated format, and the two refusals (a body that parses as nothing, and a body that is not JSON). The add endpoint suspends until the operator answers, so the approval is collected from `onAddToSchedule` and the collected request is itself the positive signal.

Driven against a real server over real HTTP, as the sibling suites are, because `start()` builds its own Netty server rather than exposing a separable Ktor module.

Suite runs in ~1s. Passes 3× consecutively with `--rerun-tasks`.

## Still open in this file

`parseRemoteItem`'s presentation-by-id branch (resolving a file hash back to a path) and the presentation upload/render helpers, which need a real deck through the rasterizer.


grunt2001c (approved):

zitlem:

- Single test file, no production change. Covers both directions across the twelve `ScheduleItem` types plus all three shapes `parseRemoteItem` accepts, including the refusal path — which is the part that would otherwise fail silently, since the DTO is one wide shape and a mis-mapped field is not a compile error.
- Real server, real HTTP, port 39_717, no collision with the existing suites.

Same non-blocking nit as #76: `kotlinx.coroutines.delay` is fully qualified in the startup poll rather than imported.

CI green.

PR #78, opened 2026-07-28
Test the Schedule tab

zitlem:
`ScheduleTabKt` was 0%-covered. 37 tests over a real `ScheduleViewModel` — the same one the app builds — so what is exercised is the wiring between the two: which view-model call each control makes, and what the list draws from the resulting state. The view model's own rules (undo history, move semantics, remote following) are already covered by the `ScheduleViewModel*` suites, so nothing here re-tests those.

## The list (9 tests)

The empty-state hint and its disappearance once there are items, the supporting detail each row type draws, row selection and what the host is told about it, and that every row offers Remove while only presentable ones offer Go Live — a label offers Edit instead.

A song row draws its number, title and songbook as three separate nodes rather than one `displayText` line (the number and songbook are what disambiguate two songs with the same title), so the ordering helper takes explicit labels rather than assuming one node per row.

## The controls (19 tests)

Move up/down including the no-op at either end, per-row and toolbar remove, clear, new, undo and redo, go-live routing, label editing, notes, and zoom.

Worth noting for review: `presentItem` is written as `handler?.invoke(item) ?: onPresenting(type)`, and `invoke` returns `Unit`, so the `onPresenting` fallback fires **only** when no handler was supplied. That is two separate tests here — a handler receives the item, and a type with no handler still gets the output switched so the operator is not left looking at the previous content.

## Dropped files (9 tests)

`handleDroppedFiles` becomes `internal` (with a comment saying why) so the drop handling can be tested directly: the drop itself arrives through an AWT `DropTarget` installed on the real window, which a headless test cannot deliver. Everything below it — classification, the view-model calls, the resulting schedule — is the real thing.

What it pins is the step after classification, where the surprises live: dropping a single image adds *its whole folder* as the picture source (a one-image slideshow would be useless), a folder with no images is ignored rather than added as an empty picture item that would present a blank screen, and an unrecognised file is skipped rather than added as the wrong type.

## Timing

Every test is ≤0.09s except the first one in each class, which pays ~1.6s of one-off Compose warm-up regardless of which test it happens to be. Full suite passes 3× consecutively with `--rerun-tasks`.

### DIFF
--- a/composeApp/src/jvmMain/kotlin/org/churchpresenter/app/churchpresenter/tabs/ScheduleTab.kt
+++ b/composeApp/src/jvmMain/kotlin/org/churchpresenter/app/churchpresenter/tabs/ScheduleTab.kt
@@ -807,7 +807,13 @@ fun ScheduleTab(
     }
 }
 
-private fun handleDroppedFiles(files: List<File>, viewModel: ScheduleViewModel) {
+/**
+ * Adds each dropped file to the schedule as whatever [classifyDroppedFile] says it is.
+ *
+ * `internal` rather than private so the drop handling can be tested directly: the drop itself
+ * arrives through an AWT `DropTarget` on the real window, which a headless test cannot deliver.
+ */
+internal fun handleDroppedFiles(files: List<File>, viewModel: ScheduleViewModel) {
     for (file in files) {
         if (file.isDirectory) {
             // Folder dropped — count image files inside and add as pictures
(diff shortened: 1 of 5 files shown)


grunt2001c (approved):

zitlem:

- The only production change is widening `handleDroppedFiles` from `private` to `internal`, which is the documented preference over reflection for reaching non-public code, and the drop itself genuinely can't be delivered headlessly (it arrives on an AWT `DropTarget` on the real window).
- The harness builds a real `ScheduleViewModel` under an isolated `user.home` and calls `TestSingletons.latchToTestHome()` first, so the JVM-wide loggers don't latch onto the temp dir that teardown deletes.
- Tests assert the schedule the operator ends up with rather than that a view-model method was called, and deliberately don't re-test the view model's own rules (already covered by the `ScheduleViewModel*` suites).

CI green, merges cleanly.

PR #79, opened 2026-07-28
Added unit tests

grunt2001c:


zitlem:
Reviewed, and pushed a fix for the CI failure (cd41688b).

**The failure.** All 8 `InstanceLinkLoggerTest` failures came from the new `OptionsContentTest`, not from the logger. That test points `user.home` at a temp dir and builds a `PresenterManager` and a `CompanionServer`, whose Instance Link paths log; `InstanceLinkLogger` resolves its log directory in a `by lazy`, so whichever test logs first in the JVM decides it for the whole run. The temp dir is deleted in teardown, every later write fails silently (the logger is best-effort), and the failure surfaced in a class that did nothing wrong.

`TestSingletons.latchToTestHome()` exists for exactly this and is already called by the `ScheduleViewModel` suites — `OptionsContentTest` was the one `user.home` swapper missing it. One call added, no production change. The pair passes three consecutive runs locally, and the run-stamped log lands in `build/test-home` again rather than in the temp dir.

**On the rest of the PR** — the extraction pattern is right and consistently applied: the `DialogWindow` call is left as the only statement the window function makes, and everything reachable moves into an `internal *Content` composable with its state hoisted. Checked the diffs for behaviour drift beyond the move and found none; `LocalLibraryDialog` goes further and pulls the entry-combining rule into `internal fun libraryEntries(...)`, which is exactly the "pull each decision into its own function" shape. Tests assert real outcomes — the values handed to `onConnect`, the settings a transform produces, which item ends up live — rather than that a stub was called. The two `mockkObject` uses (`PlanningCenterClient`, `StockMediaClient`) are both network clients and are unmocked in teardown, which is a justified last resort.

Two non-blocking notes:
- `PlanningCenterImportDialogContent` and `BibleCatalogBrowserDialogContent` take the view model (and a list of them) as a parameter. It stays inside the file that owns it, so it isn't the cross-boundary case AGENT.md forbids, and the alternative here would be ~15 typed callbacks — but it is worth being deliberate about, since it is the same shape as the documented `MainDesktop`/`CompanionSurfaceTab` deviations rather than a new pattern.
- `InstanceLinkContentTest` addresses its text fields positionally (`onAllNodes(hasSetTextAction())[0..2]`). It fails loudly rather than silently if a field is added, so it's fine, but a `testTag` would survive layout changes.

Merging once CI is green.

PR #80, opened 2026-07-28
Test the Pictures tab

zitlem:
`PicturesTabKt` was 0%-covered; this takes it to 66.6% (404 of 607 lines) with 22 tests, driven through a real `PicturesViewModel` over a real folder of real image files.

The images are written with `ImageIO` rather than as stub bytes: the view model decodes each one into a thumbnail, so a file that will not decode would exercise the error path instead of the one under test.

## The tab (13 tests)

The tab is a slideshow remote — an operator picks a folder before the service and then drives it with three buttons — so these pin that the counter, the buttons and the grid agree with each other:

- The no-folder state: a placeholder for the path, an explanatory grid, and the way out on screen.
- Loading: images only (a `.txt` in the folder is not one), sorted by name.
- The counter is one-based against a zero-based index, so it reads the way an operator would say it.
- Next/previous including wrap-around in both directions, since looping is the default and a slideshow must not dead-end mid-service.
- Play swapping to Pause and back — the same button, so both states are checked for presence *and* absence.
- Adding to the schedule hands over the whole folder even after stepping through it, rather than narrowing to the current image.

## The settings (9 tests)

Hold interval, transition duration and animation type. Each lives in two places at once — the view model drives the slideshow now, the host persists it for next time — so every test asserts **both**: a control that updates only one either fails to take effect or silently forgets itself on restart.

Also covers the clamping an operator relies on after a typo (300s → 30, 0 → 1), non-numeric input leaving the running slideshow alone, and Cancel changing nothing. The animation type is checked to round-trip as its `Constants` value rather than the translated label, and "None" specifically, since it has to persist as a real value rather than read as "never configured".

## Note for whoever runs this locally

While verifying, the whole test JVM started failing to load classes (`NoClassDefFoundError` on synthetic lambdas, every class reporting "Test process encountered an unexpected problem"). It was a corrupted incremental-compile state — an internal Kotlin compiler error had left a partial `build/classes`. Deleting that directory was not enough because the build cache restored the partial set; `./gradlew :composeApp:clean` fixed it. Same smell as the tooling issue already recorded in `DEVELOPMENT_GUIDE.md`'s decision log. Nothing to do with these tests, but worth knowing.

Full suite passes 3× consecutively with `--rerun-tasks` from a clean state. Every test ≤0.1s except the first in each class, which pays one-off Compose warm-up.

### DIFF
--- a/composeApp/src/jvmTest/kotlin/org/churchpresenter/app/churchpresenter/tabs/PicturesTabSettingsTest.kt
+++ b/composeApp/src/jvmTest/kotlin/org/churchpresenter/app/churchpresenter/tabs/PicturesTabSettingsTest.kt
@@ -1,0 +1,142 @@
+@file:OptIn(androidx.compose.ui.test.ExperimentalTestApi::class)
+
+package org.churchpresenter.app.churchpresenter.tabs
+
+import androidx.compose.ui.test.onNodeWithText
+import androidx.compose.ui.test.performClick
+import androidx.compose.ui.test.performTextReplacement
+import org.churchpresenter.app.churchpresenter.utils.Constants
+import org.churchpresenter.app.churchpresenter.models.AnimationType
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertNull
+import kotlin.test.assertTrue
+
+/**
+ * The slideshow settings on the Pictures tab: how long each image is held, how long the transition
+ * takes, and which transition it is.
+ *
+ * Each of these lives in two places at once — the view model drives the slideshow now, and the
+ * settings are persisted by the host for next time — so every test checks both, because a control
+ * that updates only one of them either fails to take effect or silently forgets itself on restart.
+ *
+ * See `PicturesTabTestSupport.kt` for the harness.
+ */
+class PicturesTabSettingsTest {
+
+    @Test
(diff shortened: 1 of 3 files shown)


zitlem:
Reviewed, and pushed a fix (13b2a586).

**The collision.** `PicturesTabTestSupport` carried its own copy of `renderedText`/`showsExactly`/`showsContainingText`. #74 hit the same clash against #78 and moved those three to `TabRenderedText.kt`; that landed on main while this PR was open, so the green run above predates it. Merging main in now would have made every call site in the package an `Overload resolution ambiguity`. This branch now takes main and uses the shared copy — all three imports it still needs (`SemanticsMatcher`, `SemanticsProperties`, `getOrNull`) remain in use for `contentDescriptions`.

**On the rest** — test-only, no production change, and the fixtures are real: images written with `ImageIO` rather than stub bytes, so the view model's thumbnail decode exercises the path under test instead of the error path, and a non-image file in the folder pins what counts as a picture. Assertions are on what the tab reported (the folderPath/folderName/imageCount triple the schedule would receive) rather than on stub invocations, and the one genuinely async thing — a drawn thumbnail — is waited for rather than assumed. No `Thread.sleep`, no mocks.

Verified locally: PicturesTab (22 tests) plus the Bible and Schedule tab suites all pass together on the merged result — 81 tests, no failures.

PR #82, opened 2026-07-28
Added Unit tests

grunt2001c:


zitlem (approved):
Approving — verified locally, no changes needed.

**What I checked.** 38 tests pass three consecutive runs on this branch. Timings are fine: 3.12s total, and everything except the first test is ≤0.2s — the 1.09s outlier is the one-off Compose warm-up that lands on whichever test runs first in a class, not a slow test. No new compiler warnings. Nothing new is declared at package top level, so there's no repeat of the `TabRenderedText` overload clash that bit the tab suites yesterday.

**The harness change is the right one.** Holding `AppSettings` in a `remember` and applying the transform means the settings tests assert the value the host would actually persist, and the dialog re-renders from it — that's what lets "the reset icon clears the QR message" and the transparent-background pair be about behaviour rather than about a callback firing.

Spot-checked the assertions against `QARemoteDialog.kt` rather than trusting them: Disable Public Access does call both `onStopTunnel()` and `onQaDisplayUrlChanged(serverUrl)` (line 363-365), so pinning both is correct and the second is the part a refactor would silently drop; `#1E1E2E` (line 584) and the reset-to-`""` (line 487) are the real values. The documented gap is honest, and narrower than it reads — `SlimSlider` itself has its own `SlimSliderTest`, so what's uncovered is only these two instances of it, not the component.

**Two nits, neither worth holding the PR for:**
- The text fields are addressed by their current *value* — `hasText("30")`, `hasText("48")`, `hasText("78")`, `hasText("100")`, and `hasText("")` for the QR message. That bakes the defaults into the selectors, and `onNode` throws when two nodes match, so the day a second field shares a default the failure surfaces in a test that isn't about that field. It fails loudly rather than silently, so it's fine as is — but a `testTag` on those fields would be sturdier, and `hasText("")` is the one most likely to go first.
- About ten of the pre-existing tests had their unused lambda parameter renamed `_` → `h` without using it, which is most of the diff noise in the first half. No warning, purely cosmetic.

Green and mergeable — say the word and I'll merge it.

PR #83, opened 2026-07-28
Test the Announcements tab

zitlem:
`AnnouncementsTabKt` was 0%-covered; this takes it to 84.2% (685 of 814 lines) with 28 tests. No production change.

The tab owns its view model privately (`remember { AnnouncementsViewModel() }`), so unlike the other tabs there is no instance to reach in from a test. That suits it: everything it does is observable from outside anyway — what it draws, the `AnnouncementsSettings` it hands back to be persisted, and what it puts on the `PresenterManager`. The settings are fed back into the tab as `MainDesktop` does, so a control's effect shows up on the next frame.

## The announcement (14 tests)

The empty state with its actions disabled, whitespace-only text still counting as nothing to show, each style toggle turning on and off independently, styles saved earlier surviving a round trip, screen position, going live, Show becoming Hide, and scheduling.

Scheduling deliberately strips the timer, so a text announcement cannot land in the service order as a countdown — that one is worth a look on review, since it is the sort of thing that would only be noticed in front of a congregation.

## The timer (14 tests)

Four modes that look alike on screen but behave differently, and the differences are what these pin:

- Only the two that count from a starting point (countdown, count-up) offer **Reset** — a specific time and a live clock track the wall clock, so there is nothing to reset them to.
- Only the two that can reach an end offer an **expiry message**.
- **Start previews**; only Go Live (and Send to Stage Monitor) marks the ticker live.
- Switching mode **stops** a running ticker, rather than leaving it counting for a mode that is no longer selected.

Nothing waits for a tick: the tab renders the timer from state the view model already holds, so the starting value, the mode switches and the buttons are all assertable on the frame after the click. What a running ticker does over time is covered by `PresenterManagerAnnouncementTimerTest`.

## Two things worth knowing for the next test in here

- The tab has **two** Go Live and two Add to Schedule buttons — one per half — and an ambiguous content-description lookup surfaces as "failed to inject mouse input" rather than as anything naming the real problem. The harness tells them apart by row position.
- Hiding only *requests* a clear (`requestClearDisplay`); `MainDesktop` is what takes the content down, so the presenting mode is still `ANNOUNCEMENTS` immediately afterwards. The request is the whole of the tab's part.

Full suite passes 3× consecutively with `--rerun-tasks`. Both classes run in well under a second.

### DIFF
--- a/composeApp/src/jvmTest/kotlin/org/churchpresenter/app/churchpresenter/tabs/AnnouncementsTabTest.kt
+++ b/composeApp/src/jvmTest/kotlin/org/churchpresenter/app/churchpresenter/tabs/AnnouncementsTabTest.kt
@@ -1,0 +1,201 @@
+@file:OptIn(androidx.compose.ui.test.ExperimentalTestApi::class)
+
+package org.churchpresenter.app.churchpresenter.tabs
+
+import androidx.compose.ui.test.assertIsEnabled
+import androidx.compose.ui.test.assertIsNotEnabled
+import androidx.compose.ui.test.performClick
+import org.churchpresenter.app.churchpresenter.data.settings.AnnouncementsSettings
+import org.churchpresenter.app.churchpresenter.presenter.Presenting
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertFalse
+import kotlin.test.assertTrue
+
+/**
+ * The announcement half of the Announcements tab: the text, how it is styled and placed, and how it
+ * reaches the screen.
+ *
+ * Everything here is asserted from outside the tab — what it draws, the settings it hands back to be
+ * persisted, and what it puts on the `PresenterManager` — because the tab owns its view model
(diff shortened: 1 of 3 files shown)


grunt2001c (approved):

zitlem:

Test-only, 28 tests across three files, following the established support-file pattern with prefixed helpers (`annButton`/`timerButton`), so nothing new lands at package top level to collide with the other tab suites — checked explicitly, since that is what bit #74/#78/#80.

The timer half is the part worth calling out: it deliberately never waits for a tick, and says so in the class doc. The tab renders from state the view model already holds, so the starting value, the mode switches and the button states are all assertable on the frame after the click; what a running ticker does over time is left to `PresenterManagerAnnouncementTimerTest`. That is the right split — a test that waited a second would be asserting on the clock.

Verified locally: merged with #84, #86 and #87 on top of main, 5525 tests, 0 failures.

PR #84, opened 2026-07-28
Test the Dictionary tab

zitlem:
`DictionaryTabKt` was 0%-covered; this takes it to 65.7% (490 of 746 lines) with 19 tests. No production change.

Built on `DictionaryFixture`'s miniature corpus — the same one the `DictionaryViewModel*` suites use — rather than the bundled ~14k entries, so every assertion is about data a reader of the test can actually see.

## Covered

- **The list** — every entry shown, and the count matching what is listed.
- **Search** — by Strong's number, by transliteration (what an operator who cannot type Greek reaches for), and by a word in the definition; no-results; and clearing to get the whole dictionary back.
- **Language filters** — Hebrew hiding Greek and the reverse, All restoring both, and the filter narrowing *together* with the search rather than either winning.
- **The detail pane** — every field and its label, the language badge, and one entry replacing another.
- **History** — Back and Forward disabled until there is somewhere to go, then walking back and forward.
- **Sending an entry on** — Go Live handing over the whole entry, Add to Schedule handing over the four fields a schedule row needs, and both disabled with nothing selected.

## The one that matters for future tests in here

The tab constructs a real `InterlinearRepository` — 4–8 MB of bundled JSON — and my first version let it. Alone, the class passed. In the full suite it caused `OutOfMemoryError`, surfacing in an unrelated Pictures test rather than in these, and the class took 16.8s.

`DictionaryViewModelInterlinearTest` already stubs the repository with `mockkConstructor` for exactly this reason; the harness now does the same. That took the class to 3.5s with every test ≤0.33s and removed the heap pressure. Anyone adding to this file should keep the stub.

Two Compose lookup notes are recorded in the harness alongside it:

- A merged node's texts stay **separate entries**, so a list row cannot be matched on the concatenation (`"G26ἀγάπηagape"`) even though `renderedText()` prints it that way. Rows are matched by two of their texts instead.
- Stubbing the repository makes the interlinear book/chapter filters appear, and their "All" chip makes a bare language-filter lookup ambiguous — so those are addressed as the topmost match.

Detail-pane assertions key on the **pronunciation**: the number, word and transliteration are on every list row too, so asserting on those would pass whether or not an entry was ever opened.

Full suite passes 3× consecutively with `--rerun-tasks`.

### DIFF
--- a/composeApp/src/jvmTest/kotlin/org/churchpresenter/app/churchpresenter/tabs/DictionaryTabTest.kt
+++ b/composeApp/src/jvmTest/kotlin/org/churchpresenter/app/churchpresenter/tabs/DictionaryTabTest.kt
@@ -1,0 +1,219 @@
+@file:OptIn(androidx.compose.ui.test.ExperimentalTestApi::class)
+
+package org.churchpresenter.app.churchpresenter.tabs
+
+import androidx.compose.ui.test.assertIsNotEnabled
+import androidx.compose.ui.test.performClick
+import org.churchpresenter.app.churchpresenter.viewmodel.DictionaryFixture
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertFalse
+import kotlin.test.assertTrue
+
+/**
+ * The Strong's dictionary tab: finding an entry, reading it, and sending it somewhere.
+ *
+ * Built on [DictionaryFixture]'s miniature corpus rather than the bundled 14k entries, so every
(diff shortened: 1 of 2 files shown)


grunt2001c (approved):

zitlem:

Test-only, 19 tests, same support-file pattern with `dict*`-prefixed helpers — no new top-level names that could collide with the sibling tab suites.

Verified locally: merged with #83, #86 and #87 on top of main, 5525 tests, 0 failures.

PR #86, opened 2026-07-28
Test the Lower Third tab

zitlem:
`LowerThirdKt` was 0%-covered; this takes it to 41.5% (277 of 668 lines) with 11 tests. No production change.

## Covered

The tab is a preset picker over a folder of Lottie animations, so: every Lottie in the folder listed and nothing else, the no-folder and empty-folder states both saying there are no presets, nothing fireable until a preset is chosen, one selection replacing another, firing the same preset twice, a Remove button per preset, and the generator still offered when the folder is empty (otherwise a new user has no way forward from this tab).

Fixtures are real Lottie files on disk — the tab reads and parses them itself, so a stub that never parses would exercise the error path instead of the one under test.

**One assertion worth a look on review:** going live hands the output the **Lottie JSON itself**, not just the preset name. That is the contract that lets the far side play without going back to disk; a name-only regression would look correct in the UI right up until an Instance Link follower or a Browser Source had nothing to render.

## Why 41.5% and not higher

Roughly half the file is the ATEM upload panel — slot enumeration, still/clip mode, aspect and capacity checks, upload progress. All of it reaches a switcher over the network. Everything decidable *before* the network is already covered by `CompanionServerLowerThirdTest` (name resolution, timing, key-target validation), so driving it here would mean either a ~5s socket timeout per test or mocking the transport to assert that a mock was called. I left it, and the test class says so.

## Also, for whoever picks up CanvasTab next

I started on `CanvasTab` before this and stopped: it calls `GraphicsEnvironment.getLocalGraphicsEnvironment().screenDevices` during composition (CanvasTab.kt:245), which throws `HeadlessException` under the test JVM — the same blocker `AGENT.md` records for `SystemSettingsTab`/`StageMonitorSettingsTab`.

It needs a one-line headless guard to become testable, but `CanvasTab.kt` is in PR #9's file set, so I did not touch it rather than create a conflict in a branch that has been open since 18 July. Worth doing once #9 lands.

Full suite passes 3× consecutively with `--rerun-tasks`. Suite runs in 0.23s.

### DIFF
--- a/composeApp/src/jvmTest/kotlin/org/churchpresenter/app/churchpresenter/tabs/LowerThirdTabTest.kt
+++ b/composeApp/src/jvmTest/kotlin/org/churchpresenter/app/churchpresenter/tabs/LowerThirdTabTest.kt
@@ -1,0 +1,136 @@
+@file:OptIn(androidx.compose.ui.test.ExperimentalTestApi::class)
+
+package org.churchpresenter.app.churchpresenter.tabs
+
+import androidx.compose.ui.test.assertIsEnabled
+import androidx.compose.ui.test.assertIsNotEnabled
+import androidx.compose.ui.test.onAllNodesWithContentDescription
+import androidx.compose.ui.test.performClick
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertFalse
+import kotlin.test.assertTrue
+
+/**
+ * The Lower Third tab: a preset picker over a folder of Lottie animations.
+ *
+ * An operator picks a preset before the service and fires it during, so what matters is that the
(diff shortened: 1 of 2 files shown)


zitlem:

Test-only, 11 tests, `lt*`-prefixed helpers, no top-level collisions.

The ~41% coverage this reaches is close to the tab's practical ceiling: roughly half the file is the ATEM upload panel (slot enumeration, still/clip mode, capacity checks, upload progress), all of which needs a real device. What is decidable before the network is already covered by `CompanionServerLowerThirdTest`.

Verified locally: merged with #83, #84 and #87 on top of main, 5525 tests, 0 failures.

PR #87, opened 2026-07-28
Added Unit tests

grunt2001c:


zitlem (approved):
Approving with a fix pushed (b21f77d6).

**The CI failure.** `PresentationRemoteContentTest > an api key too long to fit a QR code still renders the copy fallback` — `copiedText` came back null. Not a wiring problem: the copy button renders unconditionally once `qrBaseUrl` is non-empty (a null `qrBitmap` only skips the `Image`). It's layout. The test renders a 5000-character URL *above* the button, so the button's position depends on where that text wraps, which depends on font metrics.

I probed the node's bounds locally: `Rect.fromLTRB(460.0, 749.0, 565.0, 768.0)` — its bottom edge is exactly 768, the viewport height. It fits here by a single pixel. One more wrapped line on the Linux runner puts it off-screen, where the node is still in the tree and findable but the click silently does nothing, which is why the failure read as "the callback never fired" instead of as a layout problem. `performScrollTo()` before `performClick()`, per the rule the long-scroll tab tests already follow. Everything else in the PR was already green.

**On the rest.** The production changes are the extraction pattern AGENT.md asks for, not behaviour changes: pure decisions pulled into their own testable functions (`formatInstanceLinkAge`, `installerSuffixFor`, `downloadProgressFraction`, `resolveRemoteEventPresentation`, `buildContactRequest`, `statusForOutcome`) and `private` → `internal` widenings. Nice catch on renaming `InstanceLinkToast`'s `TOAST_AUTO_DISMISS_MS` — widening `RemoteActivityToast`'s copy to `internal` would otherwise have collided with it as a same-package top-level redeclaration.

The `mockk` uses are the justified kind: `Desktop` (opens a real browser/mail client), `JOptionPane` (modal), `FileChooser.Companion` (native dialog) — none reachable headless — and `unmockkAll()` in teardown is safe here because the test task doesn't set `maxParallelForks`. The assertions are on real outcomes rather than on stubs (`AboutContentTest`: 18 assertions, zero `verify`).

Verified locally: this branch merged with #83, #84 and #86 on top of main gives **5525 tests, 0 failures**.

One note, not from this PR: `PresentationRemoteDialog.kt:145` has `scrollState: androidx.compose.foundation.ScrollState = ...` — a fully-qualified type in a parameter, which DEVELOPMENT_GUIDE lists as zero-tolerance. It came in with #79, so it's pre-existing; worth a one-line cleanup sometime.

PR #89, opened 2026-07-28
Test the Canvas tab

zitlem:
`CanvasTabKt` was 0%-covered; this takes it to 76.7% (437 of 570 lines) with 13 tests. No production change.

This one is a direct follow-on from #88: the tab used to call `GraphicsEnvironment.getLocalGraphicsEnvironment().screenDevices` during composition, which threw `HeadlessException` under the test JVM. With `assignedDisplayBounds` in place it composes headless and the whole tab opened up. Thanks for putting that guard in.

## Covered

A scene is a stack of sources, and the order of that stack is what the audience sees — so these pin the scene graph each control produces:

- The no-scene empty state, and that it offers a way forward rather than just being blank.
- Creating a scene (which selects it, so sources have somewhere to go), a second scene becoming current, and removing one.
- Adding a Text, Image or Clock source from the add menu.
- The properties panel replacing its placeholder once something is selected.
- Deleting the **selected** source while leaving the rest.
- Layer reordering, including that back-then-forward returns a source to where it started.
- Nothing to schedule or show without a scene.

## Two quirks recorded in the harness

Both are documented where the next author will hit them rather than silently worked around:

- **The add-source menu can only be driven once per test.** After the first use the button keeps its hover tooltip, and a second click surfaces that instead of reopening the menu — the menu never appears and the item cannot be found. So each source type gets its own test, and tests needing several sources seed them through the view model.
- **Deleting had the same shape.** Driven immediately after the menu, the lingering popup swallowed the click. That test now seeds two sources and deletes the selected one — which also makes it a stronger assertion, since the other source has to survive.

Full suite passes 3× consecutively with `--rerun-tasks`. Suite runs in ~2.7s, all of it the class's one-off Compose warm-up.

### DIFF
--- a/composeApp/src/jvmTest/kotlin/org/churchpresenter/app/churchpresenter/tabs/CanvasTabTest.kt
+++ b/composeApp/src/jvmTest/kotlin/org/churchpresenter/app/churchpresenter/tabs/CanvasTabTest.kt
@@ -1,0 +1,169 @@
+@file:OptIn(androidx.compose.ui.test.ExperimentalTestApi::class)
+
+package org.churchpresenter.app.churchpresenter.tabs
+
+import androidx.compose.ui.test.performClick
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertFalse
+import kotlin.test.assertTrue
+
+/**
+ * The Canvas tab: the scene compositor's control surface.
+ *
+ * A scene is a stack of sources, and the order of that stack is what the audience sees — so what
+ * these pin is the scene graph each control produces: which scene is current, what is in it, and in
+ * what order. [org.churchpresenter.app.churchpresenter.viewmodel.SceneViewModel]'s own rules are
+ * covered by the `SceneViewModel*` suites; nothing here re-tests those.
+ *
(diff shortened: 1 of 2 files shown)


zitlem:

Test-only, 13 tests across a test class and a support file, with `canvas*`-prefixed helpers so nothing new lands at package top level to collide with the sibling tab suites — checked explicitly, since that is what bit #74/#78/#80.

This is the follow-on to #88, and the doc comment says so. Worth noting the two do not overlap as much as it looks: this harness composes with a default `AppSettings()`, whereas `CanvasTabSmokeTest` pins the case that was actually broken — an assignment whose stored bounds *and* index both point at screens that are not attached, which is the shape that used to reach the throwing branch. Keeping both is right; if this harness ever takes an `AppSettings` parameter, the smoke test's second case could fold in here and the file retire.

Real `SceneViewModel`, isolated `user.home` with `TestSingletons.latchToTestHome()` first, and the assertions are on the resulting scene graph rather than on stub invocations.

Verified locally: merged with #90 on top of main — 5716 tests, 0 failures.

grunt2001c (approved):

PR #90, opened 2026-07-28
Test the Crossword tab

zitlem:
`CrosswordTabKt` was 0%-covered; this takes it to 82.5% (235 of 285 lines) with 8 tests. No production change.

The puzzles are the real ones: both shipped `.xwp` files are decoded by the tab itself, so the tests load what a user would actually see rather than a fixture that could drift from the format the decoder expects. The assertions stay off any particular puzzle's words and pin the shell around them.

## Covered

- The bundled puzzles loading, and the first level opening.
- Both directions of clues listed — a crossword with only one is unsolvable.
- Checking an empty grid reporting "Wrong — try again" rather than passing it.
- Resuming at the level the player had reached.
- An unlocked level past the last puzzle falling back to the last one, rather than indexing past the end (which matters when more levels ship later).
- Saved letters restored into the grid, and a corrupt progress string ignored rather than crashing — settings are hand-editable JSON on disk, so the parser has to survive nonsense.

## Two notes for review

**Both progress tests key on cell (3,3).** My first version stored a letter at (0,0) and asserted it appeared. It doesn't — (0,0) is a *blocked* square in level 0, which renders nothing whatever is stored against it. The restore test would have failed honestly, but the corrupt-input test beside it would have **passed for the wrong reason**: asserting a letter was absent from a cell that can never show one. (3,3) is an open square in the shipped puzzle, so the negative test can actually fail if the parser lets junk through.

**Progress saving is deliberately not asserted.** The tab debounces persistence behind a 500ms `delay` before calling `onSettingsChange`, so proving a keystroke was saved would buy that assertion with half a second of wall clock — a duration rather than work. The class doc records this, along with the fix if the cover is ever wanted: `serializeInput`/`deserializeInput` are private to the file and should be widened to `internal` and tested directly instead of through a wait.

Also worth knowing: the level label is the raw zero-based index, so the first puzzle reads "Level 0". The tests pin that rather than assuming friendlier numbering.

Full suite passes 3× consecutively with `--rerun-tasks`. Suite runs in 0.14s.

### DIFF
--- a/composeApp/src/jvmTest/kotlin/org/churchpresenter/app/churchpresenter/tabs/CrosswordTabTest.kt
+++ b/composeApp/src/jvmTest/kotlin/org/churchpresenter/app/churchpresenter/tabs/CrosswordTabTest.kt
@@ -1,0 +1,130 @@
+@file:OptIn(androidx.compose.ui.test.ExperimentalTestApi::class)
+
+package org.churchpresenter.app.churchpresenter.tabs
+
+import androidx.compose.material3.MaterialTheme
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.remember
+import androidx.compose.runtime.setValue
+import androidx.compose.ui.test.ComposeUiTest
+import androidx.compose.ui.test.ExperimentalTestApi
+import androidx.compose.ui.test.onNodeWithText
+import androidx.compose.ui.test.performClick
+import androidx.compose.ui.test.runComposeUiTest
+import org.churchpresenter.app.churchpresenter.data.settings.AppSettings
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertFalse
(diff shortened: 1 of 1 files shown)


zitlem:

Test-only, 8 tests, one file, nothing at package top level.

The part worth calling out is what it declines to test. Progress saving is debounced 500ms before `onSettingsChange` fires, and the class doc says plainly that asserting it would be buying the assertion with half a second of wall clock — then points at the right fix if it ever needs cover (widen `serializeInput`/`deserializeInput` to `internal` and test them directly rather than through a wait). That is exactly the call AGENT.md asks for, and documenting the gap is better than a slow test.

Using the two real bundled `.xwp` puzzles and then asserting only on the shell around them — which level opens, both clue directions listed, checking an unfinished grid — keeps it honest without pinning any particular puzzle's words. The load wait ends on the state transition rather than on a duration.

Verified locally: merged with #89 on top of main — 5716 tests, 0 failures.

grunt2001c (approved):

PR #91, opened 2026-07-28
Add multi-translation Bible presentation

YehorAltshuler:
## Summary

- adds separate Legacy and Multi-translation Bible modes
- supports an ordered list of more than two Bible translations
- uses the first translation as the navigation and reference source
- adds translation ordering controls, typography, spacing, and optional dividers
- synchronizes the complete translation stack through Instance Link
- localizes the new controls for all supported languages

## Compatibility

Legacy mode preserves the existing Primary/Secondary workflow and Lower Third settings. Multi-translation settings are stored independently.

## Validation

- `./gradlew :composeApp:check`
- `bash cleanup_check.sh`
- `git diff --check`

All tests pass.


zitlem:
Pushed one commit to this branch (81257928) renaming the **Legacy** mode label to **Dual translation** — `maintainerCanModify` is on, and it felt better to put it inside your PR than to land it behind you after merge. Revert it if you disagree.

**Why.** "Legacy" describes how the code evolved rather than what the mode does, and reads as *deprecated* — discouraging for what is the default and the mode most churches will stay on. It was the only occurrence of that word in the whole 2,021-string catalogue. Telling detail: all 14 of your translations rendered it as *Classic* (Классический, Klassisch, Clásico…) rather than *Legacy*, so the mismatch was already visible to whoever produced them.

**What changed:**
- `values/strings.xml` — value `Legacy` → `Dual translation`, and the key `bible_translation_mode_legacy` → `bible_translation_mode_dual`. Renaming the key is free here since it was introduced by this branch and has never shipped, so there is no resource or settings migration; leaving `_legacy` in the key would mislead the next reader.
- The 14 locale values. Rather than translating afresh I mirrored the construction each file already uses for the sibling label — same noun, same case, quantifier swapped from *several* to *two* — so e.g. `Несколько переводов` → `Два перевода`, `Mehrere Übersetzungen` → `Zwei Übersetzungen`. They read as "two translations" rather than a literal "dual", which is the idiomatic form in these languages. **They still want a native check, `kk`/`be`/`et` especially.**
- `FEATURES.md`, one line.
- `BibleSettingsTabTest` now asserts *both* mode labels by name. Nothing pinned them before — the existing test only clicked "Multi-translation". I verified the new assertion fails against the old label rather than trusting it.

Left alone deliberately: `legacyTranslationList()`, the `// legacy primary/secondary` comments and the four test names containing "legacy". Those accurately describe the legacy *settings fields*, which genuinely are the old ones, and renaming internals would inflate the diff without changing anything a user sees.

**One caveat, accepted knowingly.** The secondary Bible is optional — `legacyTranslationList()` adds it only `if (secondaryBible.isNotEmpty())` — so a church running a single Bible now sees its mode labelled "Dual translation". I raised it and the maintainer chose this label anyway; flagging it here so it is on the record rather than rediscovered later.

`./gradlew compileKotlinJvm` clean, `BibleSettingsTabTest` 154 tests green.

This does not touch the other open items on this PR — the remaining machine-translated keys, the flaky `moving a multi translation to first makes it the navigation bible` test (reproduced at 1 failure in 5 isolated runs here), and the fact CI has never run because fork workflow runs need maintainer approval.

grunt2001c:
Couple of things, should we go higher than 5-6 Bibles? also on projection settings type. in content outputs it reads 2 Bibles, the wording would need to change. also instead of having segmented button switch between duel and multi Bible view would it be better to add a button to just add Bible to the end i.e. click button it reveals drop down, click button again reveals another drop down and so on. 
image

zitlem:
@grunt2001c review requested — this got large, so here is a guide.

**It is two commits, and they are meant to be read separately:**
- `02e00759` the refactor (code + tests)
- `bed4b1d3` locales only — safe to skim, but see the caveat at the end

The PR now contains three things: Yehor's original multi-translation feature, three fixes I pushed to it earlier (label rename, picker layout, a flaky test), and then a redesign that removes the mode toggle the PR introduced. @zitlem chose to keep it as one PR rather than stacking.

### What changed and why

**The mode toggle is gone.** Primary/secondary was a special case of an ordered list, so having both as switchable modes meant two sources of truth for one concept. `translations` is now authoritative, with a version-6 migration converting a pre-list file once at load.

**Read `BibleSettings.kt` first** — everything else follows from it. The old `primary*`/`secondary*` fields are deliberately left in the document so an older build still finds a configured bible on rollback, and `withTranslations` keeps their *names* in sync with the first two of the stack.

### The three things most worth your scrutiny

1. **A single translation no longer renders in the first one's styling.** The old `"secondary"` mode did `drop(1)`, *promoting* the second verse into the primary slot. Styling is per-verse now, so an output set to one translation renders in that translation's own colours. Correct, but visibly different for anyone using it.

2. **The stack is capped to what is configured** (`BiblePresenter`, `allowed`). Without it, a verse arriving from a linked instance carrying two languages drew both on a machine set up for one. `PresenterRenderTest` covers it; I broke this and the existing test caught me.

3. **Empty selection means "all"** for per-output translations, so a bible added later appears on outputs showing everything. The alternative — storing explicit indices — means revisiting every output whenever you add a translation.

### Known gaps, stated plainly

- **80 tests deleted from `BibleSettingsTabTest`.** They drive the Primary/Secondary columns this removes. That is a real coverage hole, not a tidy-up, and the class doc names exactly what is no longer covered — most importantly that editing section N writes `translations[N-1]`. Rewriting that suite is the agreed follow-up PR.
- **The 686 translations are mine, not the translation process.** Nine are copyright/licensing notices; they want a native speaker before release. Confidence is lower for kk/be/et.

Full suite locally: **5612 tests, 0 failing.** CI approval needed on this run since it is a fork PR.

YehorAltshuler:
@zitlem @grunt2001c

I prefer the unified ordered translation list over separate modes. I originally added the mode switch mainly to avoid breaking the existing primary/secondary and lower-third logic, not because I wanted two workflows permanently.

I do not use the lower-third workflow myself, but I agree it should be represented as an output-specific translation selection rather than as a special secondary-Bible rule.

For the number of translations, our church normally uses three and occasionally four. The previous application we used allowed up to seven, so I think a maximum of seven is a sensible practical limit.

grunt2001c:
> @zitlem @grunt2001c
> 
> I prefer the unified ordered translation list over separate modes. I originally added the mode switch mainly to avoid breaking the existing primary/secondary and lower-third logic, not because I wanted two workflows permanently.
> 
> I do not use the lower-third workflow myself, but I agree it should be represented as an output-specific translation selection rather than as a special secondary-Bible rule.
> 
> For the number of translations, our church normally uses three and occasionally four. The previous application we used allowed up to seven, so I think a maximum of seven is a sensible practical limit.

there are few things that i've noticed with latest build, the Bible checkbox selection does not act like a grouping checkbox, selecting/deselecting it should act as grouping, and when you deselect one of the Bibles the main Bible checkbox should become delected. 
when you add Bible in drop down it copies last Bibles setting in new drop down, should default to "Add Translation" option. the output screen preview from screen about only shows one Bible, the last one you touched. should show all selected Bibles. also need to make sure settings are backward compatible so that we don't break experience for existing users, making them re-apply new settings. also we should proabably cap max number of Bibles to 6 or 7 like you said, on current i kept going to 11.

PR #92, opened 2026-07-28
Add song number option for title slides

YehorAltshuler:
## Summary

- adds a separate option to include the song number before the generated title-slide heading
- disables that option while song title slides are turned off
- preserves the existing numbered-title behavior by default
- adds UI, formatting, persistence, and label-inventory coverage

## Why

Song title slides always concatenated the catalog number and title. This produced redundant headings when the song title already included its songbook number, for example:

`0001 – Songbook No. 1 – Song title`

Operators can now turn off the catalog-number prefix and display only the song's stored title.

## Validation

- `./gradlew :composeApp:check --no-daemon`
- changed tests passed three consecutive times with `--rerun-tasks`
- `bash cleanup_check.sh`
- `git diff --check`

The full test suite passes. Line coverage: 65.6%.


grunt2001c:
LGTM, let me know when it's no longer draft

grunt2001c (approved):

PR #103, opened 2026-07-30
Added unit tests

grunt2001c:


zitlem:
Reviewed. CI green; I also ran the 20 new suites locally three times with `--rerun-tasks` — no flakes, and every individual test is inside the ~1s bar bar one (below).

The production edits are the right shape for this repo: decisions pulled out into functions (`isVirtualizedEnvironment`, `sinkInputIndicesForProcess`, `pictureKeyAlpha`), `private` widened to `internal` rather than reached by reflection, and one injectable parameter where the unreachable step is (`windowHost`, `cefInitialized`). `pictureKeyAlpha` collapses `PicturePresenter`'s three-branch `when` into the same two-branch rule `SlidePresenter` had — behaviour is identical, `CROSSFADE` fell through to `transitionAlpha` either way. Nothing here adds a mutable seam on a singleton, and the mocked tests (`WebNavControllerTest`) assert real controller state with `verify` only in support, which is what the guidance asks for.

Merging. Four things worth a follow-up, none blocking:

1. **`cleanupLegacyJcef with no home override reads the real user home`** is safe only because `activeRoot == homeRoot` takes the early return — and it asserts nothing, so nothing would notice if that guard moved. As written, a change to that first line turns this test into a recursive delete of the developer's real `~/.churchpresenter/jcef` and `webview-cache`. Worth passing a temp `homeDir` and asserting the no-op, like the test above it does.
2. **`routeAudioToDevice("some-sink")`** actually shells out to `pactl`. It's a no-op wherever pactl is missing, but on a Linux box with PulseAudio it issues real `move-sink-input` calls.
3. **`patchJcefModuleAccess runs without throwing`** mutates the test JVM's module exports as a side effect. Harmless in practice, but it's global state changed by a test that asserts nothing.
4. **`navigating across many slides keeps only a window of the deck cached`** is 1.5s — the only one over the bar. The cost is real work (5 PDF pages rasterized), not waiting, so it's the acceptable kind; 3 pages would still exercise the window and come in under a second.

Also, the assert-nothing smoke tests (`dispose is a no-op`, `patchJcefModuleAccess runs without throwing`, `routeAudioToDevice ...`) buy coverage rather than protection — fine as a deliberate trade, just worth knowing they'd stay green through a rewrite of what they call.

PR #105, opened 2026-07-30
Fix two live bugs found while auditing untestable code, and cover ScheduleTab/PicturesTab

zitlem:
Two bugs that had been recorded as test-environment artifacts turned out to be real. Both are fixed
here, with the tests that now guard them, plus CI follow-ups and coverage for two tabs.

## 1. `NumberSettingsTextField`'s arrows could never be clicked

The component's text `Column` holds a `fillMaxWidth()` child and carried no `weight`, so in the `Row`
it was measured against the full available width and consumed all of it — the 20dp arrow column
measured `0 × 16`. Both arrows were undeliverable in the app, not only in a test, and the `range`
clamp they are the only enforcer of therefore never ran. This affects **every numeric setting in the
app**: font sizes, ports, opacities, window offsets, timer values.

Fix is one `Modifier.weight(1f)`.

**Three suites had pinned the defect in place** — `the stepper arrows are published but laid out
unusably`, in the Dictionary, Projection and Stage Monitor settings tests. Their doc comments
diagnosed the cause correctly (*"its `BasicTextField` takes `fillMaxWidth()`, leaving the arrow column
no room"*) and then asserted the breakage. They failed as soon as the layout was fixed, which is how
they were found. All three now assert the arrows are clickable and that clicking one steps a value,
through one shared helper.

The assertion is deliberately **relative** — an arrow must occupy space *exactly where its own field
does*. These tabs legitimately compose sections that are measured to nothing (a collapsed stage
monitor zone, an output row for absent hardware), and their fields measure `0x0` too; an absolute
"every arrow is non-zero" assertion would fail on those and tell you nothing about the bug.

## 2. WebSocket broadcasts sent during a client's connect snapshot were lost

`CompanionServer`'s WS handler wrote the entire connect snapshot and *then* launched its
`broadcastChannel` collector. That flow has no replay, so anything emitted in the window between was
delivered to nobody for that session and was gone — the client went on showing stale content until
its next reconnect. A phone or a linked instance connecting at the moment the operator changes the
schedule is precisely the case.

It now subscribes via `onSubscription` **before** the snapshot and holds delivery on a
`CompletableDeferred` completed after it, so a broadcast raised meanwhile queues in the flow's own
buffer instead of being dropped, and the whole snapshot still precedes any of them. A cancelled scope
releases the wait through the job's completion, leaving the previous behaviour rather than hanging.

This is also the cause of `CompanionServerTest`'s rare CI failure. That test took "first frame
received" as proof the session was registered for broadcasts; it was not. It is now.

## 3. CI

- **`push:` no longer filters on `main`.** A branch whose PR has merged stops matching
  `pull_request`, so work pushed to it afterwards ran nothing at all — eight commits went untested
  this way after #102. Concurrency is keyed on `head_ref || ref` so a PR run and its branch push run
  share one group rather than both building the same commit.
- **A `Slowest test classes` step** prints per-suite times from the JUnit XML the test step already
  writes. A previous diagnosis of the 25-minute step kill was wrong and went unchallenged largely
  because nothing recorded where the time went.
- **Budget 25 → 40 minutes** (job 45 → 60). Runs land at 11–14 minutes.

Locally the suite is 224s of test time across 411 suites, slowest `CompanionServerTest` at 13.9s.
`CompanionServerLowerThirdTest` is nowhere near the top, so its `192.0.2.1` ATEM host is **not**
costing a UDP timeout per test as had been suspected. The 25-minute kill itself remains unexplained.

## 4. Coverage — `ScheduleTab` and `PicturesTab`

- **`ScheduleTabMenuActionsTest`** drives the `ScheduleTabActions` the tab publishes via
  `onActionsReady`. Every File/Edit menu item, keyboard shortcut and approved remote request runs
  through that set, none of it touches a button in the tab, and none of it had ever been invoked by a
  test — an action bound to the wrong view-model call would look perfect on screen and still wreck a
  service order. `openSchedule`/`saveSchedule`/`saveScheduleAs` stay uncovered: all three open a
  native file chooser, which throws headless.
- **`ScheduleTabAutoRestoreTest`** covers the crash-recovery prompt, where one button keeps the
  operator's unsaved service and the other deletes it.
- **`PicturesTabKeyboardTest`** covers the arrow keys and spacebar — how the tab is actually driven
  during a service. Up/down move by a row, so the tests read the column count out of the resulting
  selection rather than assuming one.

Still uncovered in `PicturesTab`, and blocked rather than skipped: the thumbnail reorder is a
*shift*-drag, and its gesture reads `keyboardModifiers`, which a test cannot set on an injected
pointer event. The recent-folders bar is reachable only through a `private object` that latches JSON
paths off `user.home` at class-init; seeding it would mean a process-wide singleton whose first touch
decides for the whole JVM whether it points at a temp home or the developer's real one.

## Verification

`./gradlew :composeApp:check` green, including the 75% coverage gate. New and changed suites ran 3×
with `--rerun-tasks`, all green. Every new suite is under 1.3s; the largest is the known one-off
Compose warm-up on the first test in a class.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -17,10 +17,15 @@ permissions:
   contents: read
 
 concurrency:
-  # A newer push to the same branch makes an in-flight run redundant. Keyed on head_ref for a
-  # pull_request event so a branch's own push run and its PR run share one group instead of both
-  # running the same commit twice -- now that every branch push triggers this.
-  group: tests-${{ github.head_ref || github.ref }}
+  # A newer push to the same branch makes an in-flight run redundant. Keyed on the bare branch name
+  # so a branch's own push run and its PR run share one group instead of both building the same
+  # commit -- now that every branch push triggers this.
+  #
+  # ref_name, not ref: the two events spell the same branch differently. On push, head_ref is empty
+  # and ref is `refs/heads/<branch>`; on pull_request, head_ref is `<branch>`. Those are different
+  # strings, so `head_ref || ref` put the two events in different groups and ran both -- 15 minutes
+  # of runner per commit, twice. ref_name drops the prefix, so both spell it `<branch>`.
+  group: tests-${{ github.head_ref || github.ref_name }}
   cancel-in-progress: true
 
 jobs:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -3,9 +3,10 @@ name: Tests
 # The three build workflows are all workflow_dispatch-only, so until this existed nothing ran
 # automatically on a push or a pull request -- including the submodule test suites, which were
 # green but unenforced. This runs the whole suite on every push and PR.
+# Every branch, not just main: a branch whose PR has already merged stops matching `pull_request`,
+# so work pushed to it afterwards ran nothing at all until this was widened.
 on:
   push:
-    branches: [main]
   pull_request:
   workflow_dispatch:
 
@@ -16,8 +17,10 @@ permissions:
   contents: read
 
 concurrency:
-  # A newer push to the same branch makes an in-flight run redundant.
-  group: tests-${{ github.ref }}
+  # A newer push to the same branch makes an in-flight run redundant. Keyed on head_ref for a
+  # pull_request event so a branch's own push run and its PR run share one group instead of both
+  # running the same commit twice -- now that every branch push triggers this.
+  group: tests-${{ github.head_ref || github.ref }}
   cancel-in-progress: true
 
 jobs:
@@ -25,7 +28,7 @@ jobs:
     runs-on: ubuntu-latest
     # The suite takes ~14 minutes. Without a timeout a test that hangs instead of failing holds
     # the runner until GitHub's 6-hour default kills it, so the deadline is the failure signal.
-    timeout-minutes: 45
+    timeout-minutes: 60
     steps:
       - name: Checkout
         uses: actions/checkout@v4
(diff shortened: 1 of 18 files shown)


zitlem:
Reviewed, and pushed one fix onto the branch (`d666a872`) before merging.

**The collector could outlive the session it belongs to.** Subscribing before the snapshot is what stops the broadcast being dropped, but it also moves the collector's creation outside the `try`/`finally` that cancels it — the snapshot sends sit between the two. A client that vanishes mid-snapshot throws out of the handler before that `try` is entered, so `broadcastJob.cancel()` never runs and `snapshotSent` is never completed: the collector parks on `snapshotSent.await()` for as long as the server runs, holding the dead `WebSocketSession` through its own `send` closure. `scope` is the server-lifetime `CoroutineScope(SupervisorJob() + Dispatchers.IO)`, not the session's, so nothing else reaps it — one leaked coroutine per aborted connect, in exactly the reconnect churn this change is about.

It is new: before this PR the job was created *after* the snapshot, so a snapshot failure left nothing behind.

Not a delivery bug — `DROP_OLDEST` means a parked subscriber never blocks emitters, so other clients are unaffected.

Fixed by cancelling from the session job's completion, which covers every exit path including the ones before the `try`. The `finally` stays; it just no longer carries the guarantee alone.

**No test for it.** Reaching that path means making a snapshot `send` fail, and the only lever a test has — closing the client early — races the snapshot rather than landing inside it: often the frames are already buffered and nothing throws. That is recorded as a gap on `CompanionServerTest`'s doc comment instead of pinned by a test that only sometimes exercises what it claims to.

The rest stands up:

- **`Modifier.weight(1f)`** is the right fix, and deleting three tests that had pinned the breakage is the right call — the `range` clamp had never run on any numeric setting in the app. One to watch: `weight` behaves differently in a `Row` with unbounded width, so if a settings row ever ends up inside a horizontal scroller, that is where to look.
- **CI**: dropping the `main` filter with `group: tests-${{ github.head_ref || github.ref }}` does dedupe a branch push against its own PR run.
- Ran the touched suites three times with `--rerun-tasks`, then the full `:composeApp:check` with the fix in: green, coverage gate passing. Nothing new crosses 1s — the >1s entries are `CompanionServerTest`'s pre-existing ones.

PR #111, opened 2026-07-30
Added Unit tests

grunt2001c:


zitlem:
Reviewed. Verified locally: `./gradlew :composeApp:check` green, and the nine new suites run three times each with `--rerun-tasks` — no flakes.

The production edits are the shape this repo asks for: pure decisions pulled out into functions, `private` widened to `internal` rather than reached by reflection, and no mutable seam on anything process-wide. I checked the two extractions that could have changed behaviour rather than just moved it:

- **`decideTick`** is equivalent to the inline logic it replaces. `forceFullFrame` yields `DirtyRect(0, 0, width, height)`, which is exactly the `BrowserSourceFrame(0, 0, width, height, …)` the old branch built, and `computeDirtyRect` is still only reached when a previous buffer exists. `contentChanged` still gates `lastBuf`, `forceFullFrame` still gates `lastFullFrameAtMs`.
- **`copyFrameBytes`** keeps `remaining()` evaluated after `rewind()`, so the short-buffer bound is unchanged; `allocateDecodedFrame` returns the same clamped dimensions the `RV32BufferFormat` is then built from.

`EmbeddedMediaPlayer` is the justified kind of mock — it cannot be constructed without a native libvlc handle — and the tests assert real state (`confirmedPlaying`/`confirmedPaused`, the composited pixels, the copied ints) with `verify` supporting rather than standing alone. The `exactly = 1` / `exactly = 2` pairs are a genuine assertion about the reissue guard, not a stub check.

Three notes, none blocking:

1. **`build.gradle.kts` changes the `printCoverageLink` headline.** It now prints a test tally plus all six JaCoCo metrics instead of one line. Worth knowing because the coverage plan and its notes quote the old single-line form (`Coverage: 71.3% of lines (…)`) — the number is unchanged, the shape isn't. The suite-attribute regex assumes Gradle's `tests`/`skipped`/`failures`/`errors` attribute order; if that ever shifts, the tally silently disappears rather than misreporting, which is the right failure mode.
2. **`onErrorEvent does not throw`** asserts nothing — it buys coverage of a `System.err.println`, and would stay green through any rewrite of what it calls.
3. **`pausing resets the one-time volume gate…` measures 1.05s**, the only test over the bar. It is the documented one-off `mockk` instrumentation cost landing on whichever test runs first in the class, not the test itself — every other test in that suite is ≤0.02s, and it amortises across the full run. No action needed; noting it so it isn't re-diagnosed later.

`LottieFrameStreamTest`'s one `pollUntil` waits on a positive signal (`skiaBitmap.isClosed`) inside a 1s bound and fails if it never arrives, which is the pattern `AGENT.md` asks for.

Merging once CI is green on the updated branch.

PR #113, opened 2026-07-31
Preseter unitests

grunt2001c:


zitlem:
Reviewed and merging. Full `:composeApp:check` green locally (6,698 tests, 0 failed), the eight new/changed suites run three times each with `--rerun-tasks` — no flakes.

The single production change, `private val loading` → `internal val loading`, is the right seam: read-only, on an instance rather than anything process-wide, and it lets a test wait for a rasterization attempt to *finish* rather than poll `frame()`, which on the failure path never turns non-null. That is exactly the "wait for a positive signal" the guidance asks for, and `a rasterization failure is reported and leaves frame returning null` then asserts the real outcome — the cache stays empty — rather than that a call happened.

The one `Thread.sleep` is inside that bounded poll on `loading.containsKey(0)`, and the assertion below it fails if the wait ever runs out, so the timeout is the failure signal rather than the success path.

Timings are fine: nothing over the bar except `a main-frame address change is forwarded` at 1.02s, and that is position, not cost — it is the first test in its class, so it wears the one-off `mockk` instrumentation charge; the next three in the same class run at 0.02s, 0.02s and 0.00s.

Worth recording separately, because it cost me a while and will bite the next person: **on this branch a plain `./gradlew :composeApp:check` reported two `CCLIReportContentShellTest` failures and one `InterlinearRepositoryTest` failure, and a plain `:composeApp:clean` did not clear it** — a following run then failed to compile the test sources at all, with `Unresolved reference 'StageMonitorScreen'` and `initialPassClickable` for symbols that plainly exist in `jvmMain`. `--rerun-tasks` fixed it and everything passed. It is the documented corrupted-incremental-compile state, one step worse than the note in the plan says: `clean` alone was not enough, because the build cache restores the partial set. None of it was anything to do with this PR.

PR #114, opened 2026-07-31
redesign of schedule tab

grunt2001c:
Screenshot 2026-07-31 at 6 55 40 AM
Screenshot 2026-07-31 at 6 55 49 AM
Screenshot 2026-07-31 at 6 55 55 AM

### DIFF
--- a/composeApp/src/jvmMain/composeResources/drawable/ic_remove.xml
+++ b/composeApp/src/jvmMain/composeResources/drawable/ic_remove.xml
@@ -1,0 +1,9 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+    android:width="24dp"
+    android:height="24dp"
+    android:viewportWidth="24"
+    android:viewportHeight="24">
+    <path
+        android:fillColor="#FF000000"
+        android:pathData="M19,13H5v-2h14v2z"/>
+</vector>
--- a/composeApp/src/jvmMain/composeResources/values-be/strings.xml
+++ b/composeApp/src/jvmMain/composeResources/values-be/strings.xml
@@ -816,6 +816,11 @@
     <string name="settings_export_failed">Не ўдалося экспартаваць налады.</string>
     <string name="settings_import_failed">Не ўдалося імпартаваць налады. Файл можа быць пашкоджаны.</string>
     <string name="schedule_add_files">Дадаць файлы…</string>
+    <string name="schedule_item_count">%1$d элементаў</string>
+    <string name="schedule_density_compact">Кампактны</string>
+    <string name="schedule_density_normal">Звычайны</string>
+    <string name="schedule_density_detailed">Падрабязны</string>
+    <string name="schedule_kind_lower_third">Ніжняя траціна</string>
     <string name="schedule_drop_hint">Перацягніце файлы сюды, каб дадаць у праграму</string>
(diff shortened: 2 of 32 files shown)


zitlem:
Pushed `1d17d348` — the layout fixes from going through the redesign in the app, plus what came out of it.

The root cause behind two of the reported symptoms was one thing: **a row's height is never resolved.** It sits in a `LazyColumn` item, so the height constraint reaching it is unbounded and every `fillMax*` inside it silently measured to content instead. The title column wrapped its text and the Box placed it top-start — so at Compact, where the action buttons make the row taller than a single title line, the title sat visibly high. The strip's gradient is its own background, so the scrim was only as tall as the buttons: a band floating in the middle of a 54dp Detailed card with the card showing above and below it. Resolving the row's height from its children's intrinsics gives both something to fill. Worth recording that neither obvious fix works alone — `Arrangement.Center` had nothing to centre within, and `matchParentSize` on both children left nothing to size the Box.

Separately, the strip's buttons were 30dp for play against 27dp for the rest, and `Row`'s `CenterVertically` does **not** centre a mixed-height set: they came out sharing a bottom edge, leaving four of the five ~1.5dp low.

Everything else came from using it:

- **The ladder gains a rung at each end**, five in all. The original three keep their percents *and* their bands, so nothing already saved moves; the outer two borrow their neighbour's name, since they change how much room a card takes rather than what it shows. `scheduleCanZoomIn/Out` now ask the ladder for its ends instead of naming Compact and Detailed — otherwise the -/+ controls would have greyed out in the wrong place.
- **The density label reserves the width of the longest of the three names**, so the buttons beside it stop shifting as it changes. Measured rather than hardcoded, because it is "Detailliert" in German and "Подробный" in Russian.
- **A label is a heading, not a card.** It took the density's row height and full padding, so at Detailed one word sat in a 70dp band as tall as the items it heads.
- **The toolbar is one pill that wraps inside itself** rather than clipping its tail — a `FlowRow` breaks between items and never inside one, so the fixed `Row` it held could only overflow. Undo and Redo wrap as a pair, and Clear moved in with the other whole-schedule actions, which is what lets 150dp settle on two lines instead of three.
- **"Remove from Schedule" leaves the toolbar.** It acted on the current selection, so it did nothing until a row had been clicked, with nothing on the button to say so — and every row now carries its own Remove. The menu item and the Delete shortcut still reach `ScheduleTabActions.removeSelected`, unchanged; the three tests that drove the button now drive the action.
- **A saved note was drawn twice** — an italic line under the title and its own band below, both on the same `note.isNotEmpty() && !noteExpanded`.

Eleven tests across two new files pin the geometry, because every existing suite asserts on what a row *says*, and a row that says the right thing in the wrong place passes all of them. "Centred" is measured against the type chip, which is laid out `CenterVertically` and so needs no font metrics to compare against — safe across the three target platforms.

Full `:composeApp:check` green (6,733 tests), schedule suites run three times with `--rerun-tasks`, and the whole thing driven by hand in the running app.

Two things I did not touch, flagged for the record rather than as blockers: this branch also removes the Bible tab's PRIMARY badge and its test (added in #107 — that badge is what says which translation drives the book and chapter lists), and `schedule_item_count` is `"%1$d items"` in all 15 locales, so a one-item schedule reads "1 items" — and Russian needs three plural forms, not one.

PR #128, opened 2026-08-01
Expose Lottie cache internals and add server tests

grunt2001c:
Widened visibility on CompanionServer.encodeBrowserSourceFrameMessage and several LottieRenderCache members (MAX_ENTRIES, cacheDir, keyFor, cacheFile, evictOldEntries, encode/decode/scaleArgb) so tests can probe internal behavior. Expanded and added many JVM tests covering CompanionServer endpoints, InstanceLinkClient, browser-source WS/frame encoding, presentation upload/serve, planning-center OAuth, Lottie RLE/scale/eviction, AtemFrameEncoder cases, lower-third ATEM guards and sequencer auto-end. Minor test-only comment tweak in ApplyRemoteLiveStateTest. No production behavior changes intended; changes are to support thorough testing.

### DIFF
--- a/composeApp/src/jvmMain/kotlin/org/churchpresenter/app/churchpresenter/server/CompanionServer.kt
+++ b/composeApp/src/jvmMain/kotlin/org/churchpresenter/app/churchpresenter/server/CompanionServer.kt
@@ -1042,7 +1042,7 @@ class CompanionServer {
      * handled natively by the WebSocket protocol — no manual buffer/boundary parsing needed on
      * either side, and no dependence on a legacy MIME type with inconsistent engine support.
      */
-    private fun encodeBrowserSourceFrameMessage(frame: BrowserSourceFrame): ByteArray {
+    internal fun encodeBrowserSourceFrameMessage(frame: BrowserSourceFrame): ByteArray {
         val buf = java.nio.ByteBuffer.allocate(24 + frame.png.size)
         buf.putInt(frame.x)
         buf.putInt(frame.y)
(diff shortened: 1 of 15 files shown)


zitlem:
Reviewed and pushed three fixes onto this branch:

**`LottieRenderCache.cacheDir` now resolves per call rather than latching at object-init.** This one matters: `LottieRenderCache` is reachable from `CompanionServer`, `LowerThird` and `PresenterManager`, so in a full suite run something touches the object — and fixes its `cacheDir` — long before `LottieRenderCacheTest` redirects `user.home`. The eviction test then runs `listFiles { .lrcc }.forEach { it.delete() }` against whatever that latched path is. On this machine that is `~/.churchpresenter/lottie_render_cache`, 5.1 MB of real renders, and which test class happened to run first decides whether they survive. Same reasoning as `RecentLabelColors`, which resolves its file per call for exactly this.

**A guard in the eviction test** — it refuses to delete anything outside the temp home instead of discovering the problem afterwards. Verified load-bearing: re-latching `cacheDir` makes it fire (`refusing to evict outside the test home: …`) rather than delete, and the real cache was byte-for-byte identical before and after a full run of these suites.

**Restored the `CompanionServerDictionaryTest` class doc and two comments** this branch removed — including the note recording why that suite holds port 39731 (39721 is `CompanionServerQaModerationTest`'s) and why the `Res` mock does not carry across Ktor's worker threads. The merge base has those lines, so they were deleted rather than simply never written.

Everything else reads well — the widenings are all `private`→`internal` with no behaviour change, ports are distinct across the new suites, no imports left orphaned, and the `Thread.sleep`/`delay` calls are poll intervals inside bounded waits that end on a positive signal. The one 200 ms settle window (`disconnect … stops further updates`) is the documented exception where an idle window is the only terminator.

PR #131, opened 2026-08-01
Cover DeckLink parsing logic and fix JNI error handling

zitlem:
## Summary
- Extract 5 internal parsing helpers from `DeckLinkManager`'s hardware-gated methods (`parseInputModes`, `parseVideoConnections`, `parseOutputInfo`, `parseDeviceStatus`, `parseInputAudio`) and `skiaBgraToArgbPixels` from `DeckLinkComposeOutput`'s capture loop, making the decision logic testable without a DeckLink card
- Fix a latent crash: all JNI-calling methods caught `Exception`, but `UnsatisfiedLinkError` extends `Error` — a missing native symbol (confirmed for `nativeListOutputConnections` and `nativeGetDeviceStatus`, which aren't compiled into the DLL) would crash the app instead of returning the default. All catch blocks now use `Throwable`
- Add `DeckLinkHardwareTest`: 9 integration tests that exercise real hardware when present (DeckLink 4K Extreme confirmed on the dev machine) and gracefully skip on CI

Test counts: DeckLinkManagerTest 22→45, DeckLinkComposeOutputTest 8→13, DeckLinkHardwareTest 0→9. Total **+37 new tests**, all deterministic across 3 consecutive runs.

No overlap with PR #130 (ATEM socket coverage) — that PR touches `server/AtemClient.kt` and ATEM test files; this PR touches only `composables/DeckLinkIO.kt`, `presenter/DeckLinkComposeOutput.kt`, and their test files.

## Test plan
- [x] All 67 DeckLink tests pass (45 + 13 + 9)
- [x] 3 consecutive `--rerun-tasks` runs confirm no flakes
- [x] Full suite passes (unrelated pre-existing failures only)
- [x] Hardware tests exercise real DeckLink 4K Extreme on dev machine
- [x] Hardware tests skip gracefully when no DeckLink card is present (CI-safe)
- [x] `catch (Throwable)` fix prevents crash on missing `nativeListOutputConnections`/`nativeGetDeviceStatus` symbols

### DIFF
--- a/composeApp/build.gradle.kts
+++ b/composeApp/build.gradle.kts
@@ -658,6 +658,17 @@ tasks.withType<org.gradle.api.tasks.testing.Test>().configureEach {
     // as if it were a real install. An empty DSN leaves the SDK permanently disabled.
     systemProperty("sentry.dsn", "")
     systemProperty("sentry.enable-external-configuration", "false")
+
+    // Opt-in gate for DeckLinkHardwareTest, which drives a real Blackmagic card: it opens the
+    // output (pushing a frame to whatever that card is wired to), installs a JVM shutdown hook and
+    // pays driver-init time, so it must never run as part of an ordinary `check`. Off by default;
+    // enable for a deliberate hardware pass with:
+    //   ./gradlew :composeApp:jvmTest -PdecklinkHardware=true --tests '*DeckLinkHardwareTest*'
+    systemProperty(
+        "churchpresenter.decklinkHardware",
+        if (project.hasProperty("decklinkHardware")) project.property("decklinkHardware").toString() else "false"
+    )
+
     testLogging {
         events("failed")
         exceptionFormat = org.gradle.api.tasks.testing.logging.TestExceptionFormat.FULL
--- a/composeApp/src/jvmMain/kotlin/org/churchpresenter/app/churchpresenter/composables/DeckLinkIO.kt
+++ b/composeApp/src/jvmMain/kotlin/org/churchpresenter/app/churchpresenter/composables/DeckLinkIO.kt
@@ -109,7 +109,7 @@ object DeckLinkManager {
             nativeListDevices().mapIndexed { index, name ->
                 DeckLinkDevice(index, name)
             }
-        } catch (_: Exception) {
+        } catch (_: Throwable) {
             emptyList()
         }
     }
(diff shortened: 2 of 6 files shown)


grunt2001c (approved):

zitlem:
﻿Reviewed and merged with one change applied on top (1d681026).

**Kept as-is:** the `parseXxx` extraction is the right shape — each hardware-gated method is now one unreachable JNI call wrapped around testable decision logic, which is exactly what AGENT.md prescribes. 45 `DeckLinkManagerTest` tests run in 0.26s. The `catch (Throwable)` fix is a genuine latent-crash fix: `UnsatisfiedLinkError` extends `Error`, so the old `catch (Exception)` would not have caught a missing native symbol.

**Changed:** `DeckLinkHardwareTest` is now gated behind `-PdecklinkHardware=true` rather than self-skipping on `isAvailable()`.

Self-skipping was not enough, because reaching the native calls has side effects that must not happen during an ordinary `check`:

- `open` pushes a frame to whatever the card's output is wired to, and calls `registerShutdownHook()` — a permanent JVM shutdown hook that sends three more black frames and sleeps 100ms at teardown. On a machine mid-service that is a visible glitch on the program feed.
- Driver init costs far more than the ~1s per-test bar.
- Resetting the memoized `available` field coupled this suite to `DeckLinkManagerTest`, whose assertions depend on that field being latched to `false`. Correct only because the save/restore is exact and tests are single-forked.
- In exchange, on CI (`ubuntu-latest`, no `libdecklink_jni.so` in `appResources/windows`) all 9 tests returned at the guard forever, asserting nothing.

With the flag off the class is now fully inert — no reflection, no system properties, no library load, no output — so the coupling is gone. Verified both ways:

- **Off:** 9 tests, 0.02s, silent.
- **On:** `[DeckLinkHardwareTest] Found 2 device(s): [0=Intensity Shuttle, 1=DeckLink 4K Extreme]`

Also dropped an unused `assertNotNull` import, and recorded the gate in `DeckLinkManagerTest`'s doc comment.

Local run note: `DeckLinkComposeOutputTest`'s two pre-existing `runComposeUiTest` cases fail on this Windows machine with `NoClassDefFoundError: org.jetbrains.skia.Surface`, alongside `SourcePropertiesCameraTest`, which this PR does not touch — environmental, and green on CI.

PR #144, opened 2026-08-02
Unit tests5

grunt2001c:


zitlem (commented):
## Review

Reviewed the full diff and ran the four affected suites locally, repeatedly. Overall this is strong work — good doc comments that explain *why* a branch matters rather than restating the code, positive-signal waits everywhere, harnesses extended rather than duplicated, and `openAtemDialog` correctly promoted into `LowerThirdTabTestSupport.kt` instead of copied. The three production changes are all minimal, all reuse existing string resources, and all are actually exercised by the new tests.

**One blocker: a new test is flaky (~30% locally).** Details and a verified patch below.

---

### 1. `PresentationTabControlsTest` — flaky, and it takes the whole class with it

`withFakeSlides` clicks the first thumbnail after a bare `waitForIdle()`:

```kotlin
vm.slideFiles.addAll(files)
waitForIdle()
onNodeWithContentDescription("Slide 1").performClick()
```

The thumbnail carries that content description only inside `if (slide != null)` (`PresentationTab.kt:1124`) — i.e. only once the bitmap has **decoded**, which happens off the composition. `waitForIdle()` does not wait for that, so the node is sometimes not there yet and the click fails as:

```
java.lang.AssertionError: Failed to inject mouse input.
Reason: Expected exactly '1' node but could not find any node that satisfies:
        (ContentDescription = 'Slide 1' (ignoreCase: false))
    at ...PresentationTabControlsTest.withFakeSlides$lambda$0(PresentationTabControlsTest.kt:39)
```

Measured on this branch, running the four affected suites together (`*tabs.PresentationTab*`, `*tabs.BibleTab*`, `*tabs.LowerThird*`, `*tabs.AnnouncementsTab*`, `--rerun-tasks`): **failed 2 of 7 runs**. It never failed with `PresentationTabControlsTest` run alone — same load-dependent shape as the `openAtemDialog` flake this PR moves into the support file, and the same misleading symptom (the failure names whichever test happened to draw it; here it was `page down also advances to the next slide` both times, but every test in the class goes through this helper).

Suggested fix — wait for the decode itself:

```kotlin
vm.slideFiles.addAll(files)
// The thumbnail carries its content description only once its bitmap has decoded,
// which happens off the composition — a bare waitForIdle() races it, and the click
// then fails as "Failed to inject mouse input ... could not find any node".
waitUntil("the first slide thumbnail to decode", 5_000L) {
    onAllNodesWithContentDescription("Slide 1")
        .fetchSemanticsNodes(atLeastOneRootRequired = false)
        .isNotEmpty()
}
onNodeWithContentDescription("Slide 1").performClick()
```

(plus `import androidx.compose.ui.test.onAllNodesWithContentDescription`)

Verified: with the patch applied, **5/5 green** on `PresentationTabControlsTest` alone and **5/5 green** on all four suites together.

Worth checking the same pattern in the sibling suites that also render thumbnails — `PresentationTabMultiFileTest`, `PresentationTabScheduleItemTest` and `PresentationTabRealDeckTest` gate on `vm.slideFiles.isNotEmpty()` in `awaitLoaded`, which is a signal about the *view model*, not about the thumbnail being on screen. They didn't fail in 7 runs, but the gap between "slideFiles is populated" and "the thumbnail node exists" is the same gap.

### 2. `AnnouncementsTabAppearanceTest > every screen position is honored by the static preview too` — 4.6s

One test, nine `clickLabel` + recompose cycles in a single composition. Isolated it costs 4.65s; the class total is 6.8s. About 1.3s of that is the JVM-wide Compose warm-up every first test pays, so roughly **3.3s is real work — ~370ms per position click**, against 66ms for the stepper click in `the loop count can be edited directly`.

`AGENT.md`'s bar is ~1s per test, and the plan notes Compose suites cost 4–6× more on CI, so this is plausibly 15–20s there. Not a correctness problem and the cost is work rather than waiting, so I don't think it blocks — but it is now the single most expensive test in the tabs package. If each click is settling a preview animation, asserting the position→alignment mapping directly (or splitting the walk) would get the same coverage for a fraction of the time. Worth a look, or a line in the doc comment saying the cost is understood and accepted.

### 3. Duplicated helpers across the three Presentation suites

`awaitLoaded(...)` and `pdf(...)` are declared identically in `PresentationTabMultiFileTest`, `PresentationTabRealDeckTest` and `PresentationTabScheduleItemTest` (`awaitLoaded` with a 20ms poll in two of them and 50ms in the third). `PresentationTabTestSupport.kt` is right there and this PR already adds `fakeSlideFiles` to it — these two belong beside it, per `AGENT.md`'s "shared helpers live in the support file; do not redeclare them". Same story, more mildly, for `press(key)` in three files and `connectedStt` in two.

Consolidating also means the fix in §1 lands once rather than three times.

### Smaller notes

- `loopCountIncrement()` picks the increment button with the largest `boundsInRoot.top` because "Increment" is ambiguous. That works, but it is positional — if a control is ever added below the loop-count field it silently starts driving the wrong one, and the test still passes. A comment saying so is already there; a `hasAnyAncestor`-based selector would be sturdier if the ambiguity ever bites.
- `AnnouncementsTabTestSupport`: `onAddToSchedule = if (withOnAddToSchedule) { { s: AnnouncementsSettings -> reports.scheduled += s } } else null` — fine, just slightly easier to read as `.takeIf { withOnAddToSchedule }` over a named lambda.
- Nice catch giving the play/pause and loop icons real content descriptions instead of `null` — that is an accessibility improvement as well as what makes them addressable.

### Verified locally

- All four affected suites, `--rerun-tasks`: 7 runs before the patch (2 failures, both the flake above), 5 runs after (all green).
- No `mockk`/`spyk`/reflection introduced anywhere in the diff; every wait ends on a positive signal with a throwing timeout.
- Every new production symbol (`withBibleColumnWidths`, `withBibleSplitPanelWidth`, the `presentation_root` tag, the three content descriptions) is used by a test, and the descriptions reuse existing `strings.xml` entries rather than adding new ones.

PR #146, opened 2026-08-02
test: cover the follower's reference-only Bible branch

zitlem:
`ApplyRemoteLiveStateTest`'s doc comment leaves this branch out, saying it "needs a real `Bible`, which `BibleViewModel`'s own suites already build". They do build one — but none of them drives this branch, so ~50 lines of `RemoteApply.kt` had never run under a test. Same shape as #141: **a doc comment deferring coverage elsewhere is not coverage.**

## What the branch does

In `FULL_REPLICA` a follower shows the primary's wording verbatim. In `REFERENCE_ONLY` it receives only the **canonical verse code** and resolves that code against its own, independently configured Bible — so an English primary drives a Russian screen in Russian and the verse text on the wire is ignored. It is the whole point of linking congregations that do not share a language.

## Approach

The fixture is the load-bearing part: the follower's module is given a display numbering that **differs from the code** (canonical Psalm 23 sits at its own Psalm 22, and canonical John 3:16 at its own 3:15), the way a Synodal module really does. A branch that echoed the incoming numbers instead of resolving them fails here — against a same-numbering fixture it would pass and prove nothing.

8 tests: the follower's own wording, book name, chapter/verse numbers and translation label; a multi-verse range surviving (only the wording is re-resolved); and the three ways it can fail to resolve — a code this translation has no verse at (versification mismatch), no local Bible configured, and a primary that sent no code at all, which must fall through to the primary's wording rather than blanking the screen. Every failure path still switches the follower into BIBLE mode, so it moves off stale content.

No production changes.

## Cost

0.49s for the class. Three consecutive `--rerun-tasks` runs of the whole `server` package: green 3/3.

### DIFF
--- a/composeApp/src/jvmTest/kotlin/org/churchpresenter/app/churchpresenter/server/ApplyRemoteLiveStateBibleCodeTest.kt
+++ b/composeApp/src/jvmTest/kotlin/org/churchpresenter/app/churchpresenter/server/ApplyRemoteLiveStateBibleCodeTest.kt
@@ -1,0 +1,212 @@
+package org.churchpresenter.app.churchpresenter.server
+
+import kotlinx.coroutines.runBlocking
+import org.churchpresenter.app.churchpresenter.TestSingletons
+import org.churchpresenter.app.churchpresenter.data.Bible
+import org.churchpresenter.app.churchpresenter.data.SpbFixture
+import org.churchpresenter.app.churchpresenter.data.settings.BibleSyncMode
+import org.churchpresenter.app.churchpresenter.presenter.Presenting
+import org.churchpresenter.app.churchpresenter.viewmodel.InstanceLinkViewModel
+import org.churchpresenter.app.churchpresenter.viewmodel.PresenterManager
+import java.io.File
+import java.nio.file.Files
+import kotlin.test.AfterTest
+import kotlin.test.BeforeTest
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertTrue
+
+/**
+ * The follower's **reference-only** Bible branch — the one path [ApplyRemoteLiveStateTest] leaves
+ * out, saying it "needs a real `Bible`, which `BibleViewModel`'s own suites already build". They
+ * build one, but none of them drives this branch, so it has never been exercised by anything.
+ *
+ * It is the whole point of a second-campus link between congregations that do not share a language.
+ * In [BibleSyncMode.FULL_REPLICA] the follower shows the primary's wording verbatim; in
+ * [BibleSyncMode.REFERENCE_ONLY] it receives only the **canonical verse code** and resolves that
+ * code in its own, independently configured Bible — so an English primary drives a Russian screen
+ * in Russian, and the verse text on the wire is ignored.
+ *
+ * Two things make this worth its own suite rather than a happy-path check:
+ *
+ *  * **The follower's own numbering must win.** Translations disagree about where a verse lives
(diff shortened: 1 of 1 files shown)


zitlem:
## Review — approve, with two small follow-ups

Test-only (one new file, no production changes), and it closes exactly the gap it claims. I verified the premise rather than taking it on trust: `ApplyRemoteLiveStateTest`'s doc comment does defer this branch to "`BibleViewModel`'s own suites", and none of them calls `applyRemoteLiveState`, so the reference-only path really had never executed under a test.

Ran it on the branch: **8 tests, 0 failures, 0.469s** for the class — matching the stated cost.

### What makes this a good suite rather than a happy path

The fixture is the load-bearing choice and it is the right one. `SpbFixture` already separates the code reference from the display numbering (its own doc comment cites Russian Psalms following the LXX), and this suite is the first to actually use that separation: canonical Psalm 23 sits at the module's Psalm 22, canonical John 3:16 at its own 3:15. I traced `Bible.getVerseDetailsByCode` — it resolves through `codeIndex` and returns `bv.chapter`/`bv.verseNumber`, the module's own numbers — so `assertEquals(22, verse.chapter)` genuinely distinguishes resolution from an echo of the incoming reference. Against an equal-numbering fixture every one of these tests would pass while proving nothing.

The failure paths are covered where it counts, and the assertions are about consequence rather than mechanism:

- The unresolvable-code test uses book 1, which the fixture does not carry, so it walks the whole of `getVerseDetailsByCode` — `codeIndex` miss, `codeToDisplayMap` miss, `getVerseDetails` null — rather than stopping at the first branch.
- Asserting `Presenting.BIBLE` on every failure path is the right invariant: a follower that cannot resolve the verse must still move off the song it was showing.
- The "no code at all" test is the one I would most have wanted, and it is here: an older primary must fall through to the primary's wording rather than blanking an overflow room's screen.

### Follow-ups

**1. The `InstanceLinkViewModel` is never disposed.** `apply()` constructs a fresh one per call, so a class run leaks 8 of them — each holds a `HttpClient(CIO)` and a `SupervisorJob` scope that outlive the JVM's test run. Every other suite that touches one disposes it (`ApplyRemoteLiveStateRemoteFetchTest:62`, `MirroredBackgroundDownloadTest:73`, `InstanceLinkViewModelTest:81`); only `ApplyRemoteLiveStateTest` does not, and this copies that. Worth breaking the pattern rather than propagating it — and note the BIBLE branch never touches the view model at all, so one instance created once and disposed in `@AfterTest` would do.

**2. `ApplyRemoteLiveStateTest`'s doc comment is now stale**, and it is the exact comment this PR's own thesis is about. It still reads "The BIBLE reference-only branch needs a real `Bible`, which `BibleViewModel`'s own suites already build; it is left to them." Left as-is, the next person reads a pointer to suites that still do not cover it, with no mention of the suite that now does. A one-line change to point here would close the loop.

### Nits

- `codeState()` hard-codes `bookName = "Psalms"` and the Psalm 23 text for every case, so the John test reads as a Psalms payload. Harmless — reference-only ignores both by definition, which is arguably the point — but a `bookName`/`verseText` parameter would make the John test say what it means.
- `Book(19, "Псалтирь", 23)` declares 23 chapters while the module's own numbering puts that verse at chapter 22. Nothing reads the count, so it does not affect the result; 22 would be more faithful to the module being imitated.

Neither blocks merging.

grunt2001c (approved):

grunt2001c (approved):

PR #147, opened 2026-08-02
Add tab tests and media tab a11y

grunt2001c:
Add comprehensive JVM UI tests for DictionaryTab, MediaTab (playback), and many SongsTab behaviours (favorites panel, lyrics clicks, search, sorting, selection). Improve MediaTab accessibility by adding contentDescription strings and importing media_volume resource. Extend test harnesses: DictionaryTabTestSupport gains book/chapter/strongs mocks; MediaTabTestSupport adds Instance Link hooks; SongsTabTestSupport adds verseMode, presenting reporting and temp songsDir support; optional onAddToSchedule/onGoLive toggles for tests. Also exclude CrosswordTab from jacoco coverage. Small refactors to support testing.

### DIFF
--- a/composeApp/build.gradle.kts
+++ b/composeApp/build.gradle.kts
@@ -715,6 +715,8 @@ tasks.register<JacocoReport>("jacocoTestReport") {
             // `while (true) { withFrameNanos { ... } }` confetti animation loop — can't be composed
             // headless and has no terminating state to assert on.
             exclude("**/KonamiEasterEggDialogKt*")
+            // Deliberately out of scope for this coverage effort.
+            exclude("**/CrosswordTabKt*")
         }
     )
     sourceDirectories.setFrom(files("src/jvmMain/kotlin", "src/commonMain/kotlin"))
@@ -756,6 +758,8 @@ tasks.register<JacocoCoverageVerification>("jacocoTestCoverageVerification") {
             // Kept in sync with jacocoTestReport above -- both are untestable-by-construction.
             exclude("**/MacWindowActivationKt*")
             exclude("**/KonamiEasterEggDialogKt*")
+            // Kept in sync with jacocoTestReport above.
+            exclude("**/CrosswordTabKt*")
             // App entry / window wiring: `main` itself, the root composable tree and the top bar.
             // MainKt's ~200 synthetic lambda classes come along via the `$` globs.
             exclude("org/churchpresenter/app/churchpresenter/MainKt*")
--- a/composeApp/src/jvmMain/kotlin/org/churchpresenter/app/churchpresenter/tabs/MediaTab.kt
+++ b/composeApp/src/jvmMain/kotlin/org/churchpresenter/app/churchpresenter/tabs/MediaTab.kt
@@ -109,6 +109,7 @@ import churchpresenter.composeapp.generated.resources.media_select_file
 import churchpresenter.composeapp.generated.resources.media_select_to_begin
 import churchpresenter.composeapp.generated.resources.media_unmute
 import churchpresenter.composeapp.generated.resources.media_url_placeholder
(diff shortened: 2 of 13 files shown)


zitlem:
## Review — one thing to change before merge, otherwise good

Ran the full set on the branch: **192 tests across the touched tab suites, 0 failures.** The Compose tests are written the way this repo asks for — clicking controls and asserting the effect rather than asserting a render, and `waitForDefinition` waits on the reloaded text itself with the timeout only as the failure path, which is exactly right for the `Res.readBytes`-on-IO hop it is guarding.

### Please drop the `CrosswordTabKt` jacoco exclusion

This is the one I would hold the PR for. `CrosswordTab` is not uncovered — **`CrosswordTabTest` already exists** (commit `b6dfcabc`, "Test the Crossword tab"), and it ran green in my verification on this branch: **8 tests, 0.95s**. The current report puts `CrosswordTabKt` at **84% instruction / 60% branch coverage, 2,260 of 2,679 instructions covered.**

Excluding it from `jacocoTestReport` and `jacocoTestCoverageVerification` therefore does not skip measuring something unmeasured — it **deletes real, existing, passing coverage from the measurement** and orphans a live test file from the gate. Nothing then holds that suite to account: it could rot, or be deleted, and neither the report nor the verification task would notice.

Two further reasons it reads as a mistake rather than a shortcut:

1. **It moves the number the wrong way.** `CrosswordTabKt` at 84% sits well above the project's overall instruction coverage, so removing it *lowers* the reported figure. Whatever the exclusion was meant to achieve, this is not it.
2. **The justification does not match its neighbours.** Every other entry there is untestable-by-construction and says so — `MacWindowActivationKt` (native window activation), `KonamiEasterEggDialogKt` (a `while (true) { withFrameNanos { … } }` loop with no terminating state to assert on). "Deliberately out of scope for this coverage effort" is a different kind of claim: it describes the author's scope, not the code's testability. Once one exclusion means "we did not get to it", the exclusion list stops being readable as a list of things that *cannot* be covered, and the next person cannot tell the two apart.

If something in `CrosswordTab` genuinely cannot be reached, excluding that specific thing with the reason stated would be consistent with the rest of the list. Excluding the whole file, while its tests pass, is not.

### The production changes are good

The `contentDescription` additions are the right fix and correctly state-dependent — play/pause and mute/unmute follow the actual state rather than a fixed label, so a screen reader announces what the button will *do*. Using `media_volume` for the volume popup trigger is right too. No hardcoded strings, all through `stringResource`.

### Notes, not objections

- **`DictionaryTabFiltersTest :: the entry-list filter is not offered before any book has tagged data` measures 1.87s**, over the ~1s bar. I do not think it is a real cost: the test only counts semantic nodes, and this class is where mockk's one-time final-class instrumentation happened to land in my run. `AGENT.md` calls out exactly this ("~1s on first use … amortised across the suite but can push an isolated first run near the 1s bar — check it"). Worth a sanity check that the cost moves to whichever class runs first rather than staying with this one; if it moves, it is amortised and fine as-is.
- For context, the slowest test in the set is `MediaTabTest :: typing a url and loading it reaches the view model` at **5.07s** — pre-existing, not touched here, but it is now the single largest chunk of this area's runtime and is worth someone's attention separately.
- The `withOnAddToSchedule` / `withOnGoLive` flags producing `if (flag) { { … } } else null` are a little awkward to read, but they buy coverage of the tab's absent-callback branches, which is a real behaviour worth pinning. Fine as-is.
- The `InterlinearRepository` `anyConstructed` stubbing extends the harness's existing approach rather than introducing mocks where a fixture would do, so the last-resort rule is not in play here. Worth noting that `InterlinearRepository.resetForTest()` exists for the real-data path if a future test wants it.

Everything except the jacoco exclusion looks ready to me.

PR #154, opened 2026-08-03
Add UI tests for Pictures, QA and Web tabs

grunt2001c:
Add comprehensive JVM UI tests and test-support updates: new PicturesTabExtraTest, QATabDisplayAndControlsTest, QATabFileChooserTest, and WebTabBrowserBridgeTest. Extend PicturesTabTestSupport to accept PresenterManager/InstanceLink callbacks and schedule items; add a cancel case in PicturesTabSettingsTest. Enhance QATabTestSupport with labels and helpers for export/import/sort, and add narrow-layout support to WebTabTestSupport (width-constrained Box). Also add a WebTab bookmark-remove test. All changes are test-only to improve coverage of presenter interactions, file-chooser flows, and browser bridge behavior.

### DIFF
--- a/composeApp/src/jvmTest/kotlin/org/churchpresenter/app/churchpresenter/tabs/PicturesTabExtraTest.kt
+++ b/composeApp/src/jvmTest/kotlin/org/churchpresenter/app/churchpresenter/tabs/PicturesTabExtraTest.kt
@@ -1,0 +1,168 @@
+@file:OptIn(androidx.compose.ui.test.ExperimentalTestApi::class)
+
+package org.churchpresenter.app.churchpresenter.tabs
+
+import androidx.compose.material3.MaterialTheme
+import androidx.compose.ui.input.key.Key
+import androidx.compose.ui.test.onNodeWithText
+import androidx.compose.ui.test.onRoot
+import androidx.compose.ui.test.performClick
+import androidx.compose.ui.test.performKeyInput
+import androidx.compose.ui.test.pressKey
+import androidx.compose.ui.test.runComposeUiTest
+import org.churchpresenter.app.churchpresenter.models.AnimationType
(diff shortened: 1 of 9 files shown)


zitlem:
Reviewed by re-running rather than reading: the full `tabs` package **5×** with `--rerun-tasks`, all green. That is the check that matters for a Compose PR of this size — the #144 flake only appeared in 2 of 7 full-tab-set runs and never in isolation. No `Thread.sleep`/`delay`, and the per-test times are within the bar. Merging.

Two notes, neither blocking.

**1. `QATabFileChooserTest` corrects something the coverage plan had recorded as blocked.** I had `QATab`'s export/import written off as unreachable because `FileChooser.platformInstance` opens a native dialog. Subclassing `FileChooser` and standing in only the accessor works, and the plan is now wrong rather than the code. Good find.

**2. Six tests in that file assert "nothing happened" immediately after a bare `waitForIdle()`** — the cancel cases, the two failed-write cases, and the unreadable-import case. The work runs in `coroutineScope.launch` around a *suspend* chooser call, and `waitForIdle()` does not await an arbitrary coroutine. The evidence is in the same file: the positive tests use `waitUntil { dest.exists() }` / `waitUntil { qa.questions.size == 2 }`, so the author already knew the click alone is not a barrier.

These cannot fail spuriously — the asserted state is the state you start in — so they are not flaky. The cost is the opposite one: if cancelling ever *did* start clearing questions, the assertion would only catch it when the coroutine happened to win the race, which is a regression detector that works most of the time.

Cheap fix: give `FakeChooser` an invocation counter and `waitUntil { chooser.calls == 1 }` before asserting. That is a positive signal that the coroutine reached and returned from the dialog, after which the no-op branch is straight-line. I will send that as a follow-up rather than push to this branch.

PR #180, opened 2026-08-04
Add song chord editor, transposer and stage charts

grunt2001c:
Introduce inline chord support and stage-monitor charts: add SongChordPreview UI, ChordChart, ChordTransposer and ChordSheetImporter; new SongTuning model and AppSettings tuning persistence (bpm/capo). LyricSection now holds chordLines and capo; SongsViewModel parses and folds chord-only sections. StageMonitor renders chord charts with per-zone chordColor and settings. Revamp EditSongDialog to support chord preview, transposition, capo/tempo fields, section insertion and chord-sheet paste. Update SongsTab, presenter push logic, constants, strings and tests to cover tuning and editor changes. Enables transposition, insertable chord palette and better band-facing charts.


zitlem:
Flagging one thing for the repo owner before this merges — not blocking on my own authority, and not a comment on the feature itself.

**This adds 32 strings to each of the 14 non-English locale files (448 translated strings).**

`AGENT.md` rules that out about as firmly as it states anything:

> - **NEVER** add, update, or look up translations in `values-ru/`, `values-uk/`, `values-pl/`, `values-de/`, `values-be/`, `values-cs/`, `values-kk/`, or any other non-English locale file.
> - **ONLY** add new strings to the default English `values/strings.xml`.
> - Reason: translations are managed separately; machine translations cause quality issues.

Two readings, and I can't tell which applies from the diff:

1. These came from the project's managed translation process, in which case the rule is satisfied and this note is noise — say so and I'll drop it.
2. They were generated alongside the feature, in which case this is the quality problem the rule exists to prevent, and the fix is to keep `values/strings.xml` and drop the other 14 files from the PR so the managed process picks them up.

Worth an explicit answer either way, because whichever it is becomes the precedent for the next feature PR that adds UI strings. Same reasoning as the `internal var file` seam on #169: a rule that gets quietly crossed once stops being a rule.

The rest of the PR I have not reviewed in depth yet — `ChordTransposer` and `ChordSheetImporter` are pure logic and look like the parts most worth testing hard, and I'll re-run the two new suites repeatedly rather than reading them, which is how #144's real flake turned up.

PR #182, opened 2026-08-05
Extract MainDesktop logic and add comprehensive tests

grunt2001c:
Split MainDesktop into testable components: extract 50+ pure decision functions into MainDesktopLogic.kt to fix JaCoCo instrumentation (method was near 64KB JVM ceiling), extract two sidebar composables to reduce main method size, and add four test suites (482 lines) covering background media, layout state, remote content handling, and composable rendering. Also fix PresentationTabControlsTest to wait on positive signal (thumbnail) instead of idle. This improves coverage from ~0% (uninstrumentable method) to full coverage of the refactored logic.


zitlem:
Reviewed by re-running rather than by reading, since CI passing once says nothing about a 30% flake — that is how #144's real flake was found.

## No flake

- `*MainDesktop* *MainLogic* *SingleInstanceLock* *PresentationTabControls*` with `--rerun-tasks`, **5 consecutive runs**, all green (327 tests).
- The **whole `tabs` package** once with `--rerun-tasks`, green. That is the condition that matters here — the #130 and #144 flakes only appeared when neighbouring suites ran together, never in isolation.
- **Full `jvmTest`** once, green: 568 suites, 8,157 tests, 0 failures, 345s. Run because this PR rewrites live `main.kt` and `MainDesktop.kt` call sites, so the package-level run is not enough.

Timings are comfortably inside the bar: `MainDesktopComposeTest` 5.21s / 44 tests (0.12s per test), `SingleInstanceLockTest` 0.05s, `MainLogicTest` 47 tests in ~0s. A Compose suite costs ~4–6× more on CI than locally, so expect ~25–30s there — no budget concern against the 40-minute step.

`MainDesktopComposeTest` calls `TestSingletons.latchSkikoHostOs()` then `latchToTestHome()`, in that order, with a comment saying why. No `Thread.sleep`/`delay` anywhere in the new files.

## Two things worth recording

**1. The #144 decoded-bitmap fix had regressed on `main`, and this PR restores it.** `PresentationTabControlsTest.withFakeSlides` on `main` is back to a bare `waitForIdle()` before clicking `"Slide 1"`, where #144's review had put a `waitUntil` on the node — the thumbnail's `contentDescription` only exists once its bitmap has decoded off the composition. This PR re-applies it. Worth knowing that it was lost somewhere between then and now.

**2. `SingleInstanceLockTest` closes a gap that was previously written off.** The coverage notes had rejected `acquireSingleInstanceLock` on the grounds that it leaks a bound port between tests. Taking an OS-assigned free port per test and calling the new `releaseSingleInstanceLock()` in `@AfterTest` solves that properly. That "cannot be tested" entry was wrong.

## Request, not a blocker: the trivial half of the extraction

`MainLogic.kt` and `MainDesktopLogic.kt` fall **outside** the JaCoCo exclusion globs (`MainKt*`, `MainDesktopKt*`, `composeApp/build.gradle.kts:768-770`), so everything moved into them counts toward the enforced gate. That makes the shape of each extraction a coverage-integrity question rather than only a style one.

Most of them earn their place — `resolveStartupLanguage`, `withGeneratedDeviceId`, `shouldConnectCompanion`, `activeApiKey` and `browserSourceOutputAt` (both written twice before), `withAnnouncementFrom`, `advanceKeySequence`, `remoteSelectedVerses`, `toggleHiddenTabs`, `parseVerseRangeEnd`. The doc comments on several of them carry real knowledge that was nowhere in the code before; `tunnelJustDropped`'s edge-trigger note and `shouldInvalidateBackgroundCache`'s stale-cache note are both worth keeping on their own.

But a group of them are identity one-liners whose call site now reads *worse* than the expression they replaced, and which add a covered line each without adding safety:

`nullIfEmpty`, `isMediaLive`, `isPresentationLive`, `shouldConnectObs`, `isTunnelConnected`, `instanceLinkEnabledChanged`, `shouldDisconnectInstanceLink`, `isFirstEverUpdateCheck`, `hasFetchableSlide`, `shouldInvalidateBackgroundCache`, `shouldRestoreWindowGeometry`, `isValidSlideIndex`, `isUsableImageFile`.

`shouldRestoreWindowGeometry(savedPlacement == WindowPlacement.Floating, appSettings.windowX)` is the clearest case — the caller still computes half the condition, so the reader now has to hold both halves.

Two also take **function parameters**, which `AGENT.md`'s lambda-counting rule specifically warns about, since JaCoCo scores every lambda as its own method and each call site adds one:

- `deckLinkOutputCount(available: Boolean, deviceCount: () -> Int)` — one lambda
- `liveVerseCode(..., bookIdByName, codeReference)` — two

Suggestion: keep the extractions that deduplicate something or state a non-obvious rule, and inline the identity wrappers, moving any doc comment worth keeping to the call site as a comment. Nothing here is wrong, and I would not hold the PR for it.

## One small defect

`MainLogic.kt:208-210` — `isPresentationLive`'s doc comment reads "Whether **media** is what the presentation-live flag should report." Copy-paste from `isMediaLive` two functions up.

PR #202, opened 2026-08-05
Extract presentation helpers & add tests

grunt2001c:
Introduce a set of pure helper functions and constants in MainLogic.kt (remoteSongLineIndex, shouldSwitchToLyrics, shouldClearAfterLowerThird, shouldBroadcastSongSection, shouldBroadcastDisplayCleared, isControllerConnected, remoteActionType, MIN_TRANSITION_MS, modeCrossfadeDuration, isAnyScreenLockedTo, shouldFadeOnClear, fadeOutDuration, MIN_ANNOUNCEMENT_DISPLAY_MS, isFadeAnnouncement, isSlidingAnnouncement, shouldFadeOutAnnouncement, isFiniteAnnouncementLoop, announcementDisplayMs). Replace in-main logic in main.kt with these helpers and compute modeCrossfadeDuration / fade behaviour centrally. Add comprehensive unit tests in MainLogicTest.kt to cover the new behaviour (remote actions, transitions, fades, announcements, instance-link role checks). Improves clarity, testability and reduces duplicated inline logic.

### DIFF
--- a/composeApp/src/jvmMain/kotlin/org/churchpresenter/app/churchpresenter/MainLogic.kt
+++ b/composeApp/src/jvmMain/kotlin/org/churchpresenter/app/churchpresenter/MainLogic.kt
@@ -1,16 +1,24 @@
 package org.churchpresenter.app.churchpresenter
 
 import org.churchpresenter.app.churchpresenter.data.settings.CompanionSatelliteSettings
+import org.churchpresenter.app.churchpresenter.data.settings.InstanceLinkRole
 import org.churchpresenter.app.churchpresenter.data.settings.InstanceLinkSettings
 import org.churchpresenter.app.churchpresenter.data.settings.AppSettings
+import org.churchpresenter.app.churchpresenter.dialogs.RemoteEventType
+import org.churchpresenter.app.churchpresenter.utils.Constants
 import org.churchpresenter.app.churchpresenter.data.settings.BackgroundSettings
+import org.churchpresenter.app.churchpresenter.data.settings.BibleSettings
+import org.churchpresenter.app.churchpresenter.data.settings.SongSettings
 import org.churchpresenter.app.churchpresenter.data.settings.OBSSettings
 import org.churchpresenter.app.churchpresenter.utils.UpdateCheckResult
 import org.churchpresenter.app.churchpresenter.data.settings.ScreenAssignment
 import org.churchpresenter.app.churchpresenter.data.settings.ServerSettings
 import org.churchpresenter.app.churchpresenter.presenter.Presenting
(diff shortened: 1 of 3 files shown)


zitlem (approved):
Approved. Validated by re-running rather than reading, per the treatment #182 got.

**Full suite ×3 with `--rerun-tasks`, on the merged head (which already contains #201 and #203): 8,326 tests, 0 failures, 0 errors each time.** `MainLogicTest` is 158 tests in **0.014s** — pure functions, no coroutine, no Compose clock, no filesystem, so there is nothing in it that can race. That is the structural reason it is flake-free, not just three green runs.

**Extraction fidelity — checked on the production diff, which is the part worth reading.** Every helper I compared is a faithful move: `lottieProgressAt` is character-for-character the old local function, `modeCrossfadeDuration` / `fadeOutDuration` keep `MIN_TRANSITION_MS = 100` matching the old inline `.coerceAtLeast(100)`, and `announcementDisplayMs` correctly folds in the `* loopCount` that used to sit at the call site. `startupWindowPosition`/`startupWindowSize` preserve the asymmetry that was easy to lose — position is gated on `shouldRestoreWindowGeometry(isFloating, savedX)` while size is gated on `isFloating` alone.

**Mutation-checked, because 158 tests over one-line helpers invites the question of whether they assert anything.** Two mutations to the helpers with real branching, each caught by exactly one test:

- `lottieProgressAt`: `coerceIn(0f, pauseFrame)` → `coerceIn(0f, 1f)` in the pre-hold stretch → fails *"the stretch before a hold never overruns the held frame"*.
- `announcementDisplayMs`: drop `* loopCount` → fails *"each loop adds its own time on screen"*.

So the tests bite. `MainLogic.kt` lands at **100% (119/119)**, and it is inside the gate's denominator while `main.kt` is not — which makes this the honest kind of coverage gain rather than a threshold move.

**One observation, not a blocker and explicitly not a merge condition.** A number of the extracted predicates are bare restatements of their body — `shouldSwitchToLyrics(mode)` is `mode != Presenting.LYRICS`, `shouldBroadcastDisplayCleared(mode)` is `mode == Presenting.NONE`, `canPushToSchedule(link)` is `link.allowPushToSchedule`. Those add a name and a test without adding a rule. The ones that earn their place are the ones whose doc comment states something the code cannot — `shouldFadeOnClear` explaining why a locked screen must not fade, `primaryOutputScreenIndex` explaining the three fallbacks in order of trust, `lottieProgressAt` explaining why the post-hold stretch is re-scaled rather than resumed. Worth aiming the next extraction at that second kind. Genuinely useful here regardless: `isSongLineMode`, `isUsableOutputDir` and `remoteEventTargetsClient` each replace two-to-four duplicated copies, where drift between the copies was a real possibility.

PR #203, opened 2026-08-05
Support multi-language song section headers

grunt2001c:
Add a SECTION_WORDS map and use it in sectionKindOf to recognise chorus/bridge/tag headers in other languages (Russian/Ukrainian), matching ChordSheetImporter canonicalisation. Keep unrecognised labels (and Pre-Chorus) as VERSE. Update unit tests to assert multilingual matches. Also make the SongsTab title slide use the same SectionLabelRow chip for visual consistency and remove the extra divider. This unifies parsing and presentation of section headers across localized song files.

### DIFF
--- a/composeApp/src/jvmMain/kotlin/org/churchpresenter/app/churchpresenter/composables/SongChordPreview.kt
+++ b/composeApp/src/jvmMain/kotlin/org/churchpresenter/app/churchpresenter/composables/SongChordPreview.kt
@@ -50,6 +50,8 @@ import churchpresenter.composeapp.generated.resources.song_transpose_up
 import org.churchpresenter.app.churchpresenter.models.LyricSection
 import org.churchpresenter.app.churchpresenter.utils.ChordSegment
 import org.churchpresenter.app.churchpresenter.utils.ChordTransposer
+import org.churchpresenter.app.churchpresenter.utils.SongSectionWordGroup
+import org.churchpresenter.app.churchpresenter.utils.SongSectionWords
 import org.jetbrains.compose.resources.stringResource
 
 /** How a section reads in the preview — the colour tells verses from choruses at a glance. */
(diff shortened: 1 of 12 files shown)


zitlem:
Reviewed by re-running. **Six existing tests fail, deterministically — 3 of 3 runs, same six each time, so this is a straight break rather than a flake.** Your CI was still pending when I ran, so this may just be a heads-up with the diagnosis attached.

## The failures

```
SongsTabChartAndTitleSlideTest
  the title slide's own sections lead the song's
  choosing the title slide sends it out ahead of the song
  the title slide is offered when it is turned on
  going live from the title slide reports the song as presenting
SongsTabLyricsClickTest
  double-clicking the title slide sends it live
  clicking the title slide selects it and previews it, without going live
```

Reproduce with:

```
./gradlew :composeApp:jvmTest --tests '*SongsTabChartAndTitleSlide*' --tests '*SongsTabLyricsClick*' --rerun-tasks
```

## Cause

`SectionLabelRow` renders through `ZoneLabel`, which does `text = text.uppercase()`. So swapping the title slide's plain `Text(stringResource(song_title_slide))` for `SectionLabelRow(label = …)` changes the rendered string from `Song Title Slide` to **`SONG TITLE SLIDE`**.

The six tests select that row by its text, case-sensitively:

```
Reason: Expected exactly '1' node but could not find any node that satisfies:
        (Text + InputText + EditableText contains 'Song Title Slide' (ignoreCase: false))
```

The rendered tree in the third failure confirms it — `SONG TITLE SLIDE` is present, next to `VERSE 1` and `CHORUS`, which the existing section labels already uppercase the same way.

Two ways out, both yours to pick: update those six selectors to the uppercase form (consistent with how the other section labels are already matched), or give the shared test support a constant for it so the next label change moves one line. I have not touched your branch.

## The rest of the change reads correctly

I checked the part that worried me first, and it is fine:

- **Section *parsing* is unaffected.** `isChorusHeader` keys the chorus off braces `{…}`, not off any word, so a `{Припев}` was already a chorus and the auto-repeat behaviour is untouched. `sectionKindOf` only colours the chart.
- **The word lists genuinely do match `ChordSheetImporter`'s**, as your doc comment claims — I compared all three kinds term for term (chorus/refrain/припев/приспів, bridge/бридж/мост/міст, tag/outro/ending/кода/концовка/закінчення). Worth stating since two multilingual lists that must agree is exactly the kind of thing that drifts.
- **`Pre-Chorus` → VERSE is right and is pinned**, in both languages. `startsWith` means `предприпев` cannot match `припев`, which is the subtle half.

## One stale comment, in a file this PR does not touch

`ChordSheetImporter.kt`'s own doc comment says:

> `isChorusHeader` keys the chorus off braces, and the preview colours sections by **the English words** in `sectionKindOf`.

That second clause stops being true with this change. Worth updating in the same PR while the reason is fresh — it is the sort of line that later gets read as authoritative.

Issue #4, opened 2026-05-16
Where are the Bibles? Where is the documentation on how to use the app


nickmuteti:
**Describe the bug**

**To Reproduce**

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

zitlem:
Our official website is [ChurchPresenter](https://churchpresenter.org). We are currently working on more tutorials and documentation on how to use the app.

Bible translations are typically protected by copyright, so the app does not include Bible text directly. Instead, it provides a conversion tool (Help → Convert) that lets users convert compatible [Zefania XML Bibles](https://github.com/biblenerd/Zefania-XML-Preservation) for personal use.

Since many of our users come from  [SoftProjector](https://github.com/SoftProjector/softprojector), this also makes it easy to continue using Bible files they may already have downloaded there.

nickmuteti:
The app is great, but finding basic things like bibles is a night mere. Can you make it more straight forward. Maybe you can get inspiration from bibleshow. 

Since it is open Source, and non proprietary, you can support all bibles here: https://github.com/Beblia/Holy-Bible-XML-Format/tree/master . Finally have a preview  box that shows what is live. and a small stand alone output window that can be shared  on google or zoom meet. The project is a highly promising one.

grunt2001c:
We'll see about making some more videos on how to add Bibles and other things. We try to not stealing UI and ideas from other apps. but if you have suggestion let us know. We've added format of the Bibles you've linked into our converter so next release it should work

grunt2001c:
isses shuld be addressed in https://github.com/ChurchPresenter/ChurchPresenter/releases/tag/v26.3.57

grunt2001c:
here is detailed post about initial setup https://churchpresenter.org/blog/how-to-set-up-worship-presentation-software/

grunt2001c:
Holy bible xml will now be part of download bible dialog

PR #209, opened 2026-08-06
Bump ChurchPresenter-BLE: book-alias fixes, test hygiene, no service dates

zitlem:
Submodule pointer only — `1c85892 → 403bdb0`. No other file changes.

Pairs with **ChurchPresenter/ChurchPresenter-BLE#4**; merge that first.

### What the bump brings

**Three book-alias gaps**, two of which made a book effectively undetectable:

- Ezekiel's only Russian alias was the one-и typo `иезекиль`, so the correct «Иезекииль» and every inflection resolved to *nothing*. A book that never fires produces no wrong detection, so no session log would have surfaced it.
- «Второзакония» — the standard Russian title of Deuteronomy — was swallowed by the number lexicon (stem `втор` = 2 plus a valid ending) and never reached the resolver. Fixed as longest-match-wins between the two lexicons, so the next such name needs no new data.
- The canonical genitive titles («Книга Иова/Руфи/Ионы», «Тита») were missing; all ≤4 chars, so they inherit the short-alias gate and stay refused in bare prose.

**Wrong-book detection**: multi-word names now join by per-word stem, and four narration-vs-citation gates were tightened — including `judas`, the German alias for Jude, which fired on the English word for the betrayer in essentially every Passion sermon. Replay of two archived services: wrong-book emissions **4 → 1**, correct-book coverage **8 → 9**, matched go-lives and FN unchanged.

**Tests**: watcher inputs are now the shortest phrasing that reaches the gate rather than verbatim sermon sentences, verified by mutation rather than by reading — two shortenings looked correct, passed, and had silently stopped exercising their gate. Two coverage holes closed, including one gate with no effective test at all.

**Privacy**: the submodule no longer commits service dates, ids or timestamps.

### Risk

No app-side code changes. The engine is exercised by the submodule's own suite (282 tests, green on three consecutive `--rerun-tasks` runs) and `./gradlew compileKotlinJvm` passes at the repo root.

### DIFF
--- a/DEVELOPMENT_GUIDE.md
+++ b/DEVELOPMENT_GUIDE.md
@@ -216,7 +216,7 @@ object Constants {
 | Wildcard imports | 0 | — | All expanded to explicit imports ✅ |
 | Debug prints | ~8 | VideoPlayer.kt, LowerThirdSettingsTab.kt, WebsitePresenter.kt | All `System.err.println` for error diagnostics — kept intentionally |
 | Non-null assertions (`!!`) | 0 | — | Replaced with smart casts and `getValue()` ✅ |
-| Hardcoded UI strings | 0 | — | "OK" moved to strings.xml; emojis kept (not translatable) ✅ |
+| Hardcoded UI strings | 1, and it is fine | `ProjectionSettingsTab.kt:1246` | `Text("$w×$h")` renders `1920×1080` — digits and a multiplication sign, identical in every locale. **Re-audited 2026-08-06 with the corrected grep below; three real violations found and fixed since this row last said "0"** — `WebTab.kt` ×2 (`0375f909`) and `BibleTab.kt`'s Ctrl/Shift selection hint. All three were **instructions**, the only sentence explaining how something works, rather than labels. |
 | Fully qualified `androidx.compose.*` | 0 | — | 20 sites removed 2026-07-28; the verification grep below returns nothing ✅ |
 | Fully qualified `java.*` where the import exists | 0 | — | 27 redundant sites removed 2026-07-28 ✅ |
 | Fully qualified `java.*` with no import | ~164 | Widespread | **Kept deliberately** — see the decision log |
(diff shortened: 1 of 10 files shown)


zitlem:
Reviewed by running it rather than reading it, plus one content check. **The engine change is sound; I am not merging yet because the anonymisation pass damaged prose the docs depend on.**

## What passes

- **BLE module tests: 282, green on 3 consecutive `--rerun-tasks` runs.** Deterministic.
- **Both builds, per AGENT.md**: `./gradlew build --rerun-tasks` in the module and `./gradlew compileKotlinJvm` at the repo root. Zero errors from either. (The main build is the permissive one, so the module's own build passing is the load-bearing half.)

## The privacy claim holds — I checked the crux rather than the description

`SESSIONS.md` holds the pseudonym→session-id mapping, so the whole change rests on it staying out of the repo. It is gitignored (`.gitignore:11`), untracked, and **never committed on any branch** (`git log --all -- SESSIONS.md` is empty). No ISO dates remain in tracked Kotlin. That part is done properly.

## What I would fix first — the substitution was mechanical and broke text

Replacing every date with the literal string `an earlier pass` (43 sites) leaves three kinds of damage:

**1. A mangled leftover.** `TRAINING_PLAN.md:243` reads `94 emissions/0 TP in an earlier pass-08` — the substitution ate the date and left its trailing `-08` attached.

**2. Ungrammatical text**, e.g. `Config.kt:90` `// Retired an earlier pass: revisiting an EARLIER chapter...`, and in `Config.kt` around l.88-91 the replacement runs across a line break to give `…a single true positive; the / an earlier pass replay had already shown…`. Also `ContinuationEngineTest.kt:85`.

**3. The one that actually costs something — "since/before X" claims whose X is now gone:**

- `TRAINING_PLAN.md:344` — **"When triaging a recording made before an earlier pass: its logs are display-numbered — Psalms off by …"**. This is an *operational instruction*: the reader is told to treat a recording differently depending on which side of a boundary it falls, and the boundary is exactly what was removed. It cannot be followed now.
- Same shape at `TRAINING_PLAN.md:169`, `:244`, `:266` (`OFF since an earlier pass`), `ContinuationEngine.kt:45`, `tools/triage_report.py:127`.

The fix is not to put the dates back — the change is right. It is that a **relative** marker cannot express an ordering. The commit already introduces a pseudonym scheme (`S07`); the same idea covers these if the boundary is named by something monotonic and non-identifying — a schema/config version, a release, or a commit — so `before an earlier pass` becomes e.g. `before the canonical-numbering change`. Several sites already have a natural name available: the numbering switch, and `Config.chapterHistoryEnabled` going false.

Nothing here affects behaviour, and 282 tests and both builds are green — so this is a "worth one more commit before it lands" rather than a defect in the engine work. Happy to merge as-is if you would rather take it now and fix the prose separately; say the word.

PR #221, opened 2026-08-06
Extract command runner + add tests for device/window logic

grunt2001c:
Introduce a reusable CommandRunner and readCommandOutput (CommandOutput.kt) and refactor OS-specific command calls into testable functions (linuxWindowBoundsFrom, macWindowBoundsFrom, parseXwininfoBounds, cameraDevicesFor, cameraFormatsFor, vlcInstalledOn, detectVlcInstallPathFor, parseCdpMessage, etc.). Replace many inline ProcessBuilder usages with calls through a runner so behaviors can be unit-tested. Add FakeCommandRunner and a large suite of JVM tests (DeviceEnumerationTest, CommandOutputTest, WindowBoundsTest, CdpMessageTest, VlcAvailabilityTest, SharedCameraFrameCacheTest updates, and UI test additions) covering command handling, camera format discovery, window enumeration/bounds, CDP message parsing and VLC detection. Also tidy up SharedBrowserFrameCache message handling to use the new parser. Changes are geared to make platform tooling logic deterministic and thoroughly tested without changing runtime behavior.


zitlem:
Reviewed by re-running rather than reading, per the usual approach here.

**Verification**
- `composables` + `tabs` packages, 3× with `--rerun-tasks`: green every time. `data` package 3× after the Bible.kt commit landed: green. Full suite green.
- All 142 new tests are comfortably inside the 1s budget — slowest new class is `LabeledControlsTest` at 0.29s.
- Mutation-checked two of the claims the docs make: dropping the `w > 0 && h > 0` guard in `parseXwininfoBounds`, and dropping the `wmctrl.exitCode == 0` check in `linuxWindowsFrom`. Four tests failed across the two, naming the behaviour they lost. The tests are real.
- `FakeCommandRunner` is a plain fake asserting returned values and command sequences — no mocks, and it records the call order, which is the only way several of these fallback paths are observable.

**One finding, fixed on the branch (38a2fa5d)**

Six private per-OS wrappers went unreachable when `cameraFormatsFor`, `cameraDevicesFor` and `openWindowsFor` started calling the `*From` functions directly: `listDshowFormats`, `listV4l2Formats`, `listAvfoundationFormats`, `listWindowsCameras`, `listMacCameras`, `listMacWindows`. Nothing referenced them — fifteen lines of dead, uncovered code in a coverage PR. Removed. The compiler doesn't flag these (warnings look suppressed in this build), so grep is the only thing that catches them.

**On the Bible.kt commit**: checked that `conn` and `bibleId` were never assigned anywhere in the class, so the JDBC branch in `getBooks()` was unreachable and its removal is behaviour-preserving. `JdbcDatabase`/`DatabaseResult` are still live via `Songs.kt`.

The seam extraction itself is the right shape — one unreachable step taken as a parameter, the sequence around it exercised for real.

PR #222, opened 2026-08-06
Catch dialogs that open with content cut off

zitlem:
Six dialogs declare a fixed size and hold no scroll container, so content needing more room than the declared height is simply not drawn — no scrollbar, nothing on screen to say anything is missing. Nothing in the suite could catch that, because `boundsInRoot` is already clipped: a node pushed past its container reports empty bounds rather than out-of-range ones, so the usual `width > 0f` check reads "pushed off screen" and "collapsed to nothing" as the same thing.

## The bug this found

`MemoryMonitorWindow` was already in that state. Its readings measure **414dp in a 460x440 window** — 6% of slack — and **466dp** once text grows 30%, so it opened clipped for anyone running OS font scaling above ~1.06x. Raised to 500dp and given a scroll container as a floor under that. It is the only production change here.

## Why minimum intrinsic height

Two more obvious approaches both pass on layouts that are visibly broken. Each was verified against `AddLabelDialog` at the 400dp its own height comment records as broken:

- **Compose into a box of the declared size, look for nodes outside it.** `Modifier.size` caps the content and Compose honours the cap — children are measured against what's left rather than spilling over, so every node reports as inside the box while the text within is squeezed and clipped. Nothing is ever outside the box, which is precisely why the defect is invisible on screen too.
- **Compose at the real width with unbounded height.** A child with `Modifier.weight(1f)` — which the label editor's content column has — is dividing up leftover space, and infinite height cannot be divided, so it collapses. Reported 210dp for content that needs 474dp.

`AddLabelDialogHeightTest` is the control, pinning the 400dp regression as still detectable. It exists because both mistakes above looked exactly like a suite of passing tests.

## Text growth

Three of the bodies draw nothing but fixed string resources, so there is no parameter to pass a longer string through. Their growth case scales the font 1.3x instead. This avoids reading the non-English locale files (`AGENT.md` — managed separately), and avoids gating CI on translation content, where a translator lengthening a sentence would break the build. It is also a real scenario on its own: desktop OS font scaling.

## What this does not catch

The declared sizes are inline dp literals in the composables, so the tests hold a duplicate of them. This catches **content outgrowing a window** — a new row, a longer string, a bigger font — not someone editing a dp literal downward. Deliberate: growth is what happens by accident.

The colour picker overlay on `AddLabelDialog` is not measured. `ColorPickerDialog` is a Compose `Dialog`, so it composes into its own layer and contributes nothing to the host's layout (verified: 474dp with it open and closed alike). It also grew a scrollable middle section since that comment was written, so it can no longer be cut off.

## Verification

- `:composeApp:check` green.
- The new tests run 3x with `--rerun-tasks`, deterministic.
- Slowest is 0.9s (first-composition warm-up); the rest are under 0.12s.


grunt2001c (approved):

zitlem:
Reviewed. Merged current `main` in first (the branch predated #221 and #223) and validated what will actually land.

**Verification**
- The five new/changed suites 3× with `--rerun-tasks`: green every time. Full suite green, twice.
- Test cost is fine — slowest individual case is 0.72s (`the label editor fits the 640dp it was raised to`), everything else well under.
- Mutation-checked the three central claims. Two were killed cleanly: dropping the clamp in `startupWindowSize` failed three `MainLogicTest` cases by name, and making `dialogSizeWithin` return the preferred size failed two `WindowUtilsTest` cases.

**The third mutation survived, and that was the finding.** Putting `MemoryMonitorWindow` back to `height = 440.dp` — the exact regression this PR fixes — left the entire suite green.

The reason is that the test spelled the size out as its own literal, so it held the *content* to 500dp while nothing held the *window* to 500dp. The class doc recorded this as a deliberate trade ("the growth is what happens by accident, and shrinking a window is not"), but the comment added to `MemoryMonitorWindow` claimed the opposite — that `DialogViewportTest` holds the height to the measurement — and that was the one that wasn't true. It also applied to all six dialogs, not just this one.

**Fixed in ae905a5f.** `DialogSizes.kt` names each declared size once; the composable and the test both read it. Re-ran the mutation to confirm the gap is closed — `MEMORY_MONITOR_WINDOW_HEIGHT` at 440dp now fails `MemoryMonitor content still fits when its text grows`. Updated the class doc, which no longer describes the behaviour, and the `MemoryMonitorWindow` comment, which now points at where the rationale lives.

The production fixes themselves are good — the 1366x768 settings-dialog case and the drag-to-nothing window case are both real, and the reasoning is written down at each site.

Minor, non-blocking: this adds a third tracked `.impeccable/hook.cache.json` (contents `{"version":1,"sessions":{}}`). Two are already tracked on `main`, so it is pre-existing rather than new here, but the pattern is probably worth a `.gitignore` line at some point.

PR #224, opened 2026-08-07
Add tests: chord parsing, autofit, livemap

grunt2001c:
Expand unit coverage for parsing and layout edge cases. Tests assert handling of mismatched/unclosed chorus headers, blank-line and end-of-sheet chord behaviors, and heading adjacency in ChordSheetImporter. Add chord-chart font-size invariants and scenarios to AutoFitUtilsTest (cap/floor, line-count, info-line and intro-only cases). Add LiveMapReporter test to omit empty provenance fields. These tests lock down recent bug fixes and guard regressions in parsing, layout sizing and telemetry URL generation.


zitlem:
Reviewed. Merged current `main` in first, then re-ran.

**Verification**
- All five suites 3× with `--rerun-tasks`: green every time. Full suite green.
- Tests-only PR, so no production diff to read — the review is whether the tests pin anything, which means mutating the code they cover.

**Two tests named a behaviour but did not require it. Fixed in 2aa66b7d.**

`the info line costs height, so it cannot make the chart bigger` and `a chord-only intro line is one row, not a stacked pair` each asserted a non-strict inequality at a single zone height. I mutated `calculateChordChartFontSize` two ways:

- dropped the info row from the height model (`var height = 0f` regardless of `hasInfoLine`) — **suite stayed green**
- collapsed the chord-only branch into the stacked one (`if (false)` on the `words.isBlank()` check) — **suite stayed green**

Both survive because `<=` and `>=` are satisfied by the cost not existing at all. The test names the behaviour in its title and then permits its absence.

The fix sweeps zone height (100..600 step 20) and asserts both directions: never inverts anywhere, and strictly differs somewhere. That keeps it independent of platform font metrics — which is presumably why a single height was written non-strict to begin with, and I checked that a plain strict assertion at one height would have been the fragile fix. Both mutations now fail, each killed by exactly the test that names it.

The rest is good, and the invariant-over-pixel-value approach is right for this function. The two absolute-value tests (`fits its cap when there is room`, `keeps the requested size when there is nothing to draw`) are what stop a constant-returning implementation passing the monotonicity ones — worth keeping in mind as the pattern here.

Minor, non-blocking: `dragging the divider right widens the column and saves the width` measures 1.23s, just over the ~1s bar in AGENT.md. It looks like Compose first-use warm-up landing on whichever case runs first rather than the test itself, and it is well inside what existing Compose tests here cost (`MediaTabTest` is 5s), so I have not changed it.

PR #225, opened 2026-08-07
Add visual regression testing for presenter output

grunt2001c:
Introduce Roborazzi screenshot testing framework to capture what the congregation sees on presenter screens. Includes:

- PresenterScreenshotTest with 6 baseline screenshots covering songs and Bible verses in fullscreen, lower-third, and multi-language modes
- GitHub Actions workflow (screenshots.yml) that renders images on Linux and posts visual diffs as PR comments using reg-viz
- Gradle configuration for Roborazzi with output committed to composeApp/screenshots/ so changes are reviewable in diffs
- Documentation in AGENT.md explaining the workflow and platform-specific rendering considerations

Visual testing is advisory (not required for merge) since platform differences in font rendering make absolute pixel comparison unreliable. CI sidesteps this by rendering both base and current branch on the same runner.


zitlem:
Reviewed. The design here is right, and it is the part I checked hardest: rendering both sides on one runner is the correct answer to Skia's per-platform text rasterisation, and keeping the check advisory rather than required is the correct answer to "a deliberate design change looks exactly like a regression". Both are reasoned out at the call sites.

**The `screenshots` check was failing, and it would have failed on every PR.** Fixed in f3fb1ece.

```
##[error]'image-directory-path' is not directory. Please specify path to image directory.
```

The workflow passes `image-directory-path: ./composeApp/build/outputs/roborazzi`, but `build.gradle.kts` overrides `roborazzi.outputDir` to `composeApp/screenshots`. I confirmed locally that `composeApp/build/outputs` is never created at all — not by `recordRoborazziJvm`, not by a plain test run. So the step could never have succeeded, on this PR or any later one.

Changed it to `./composeApp/screenshots`, which is where the record step writes and therefore what reg-actions needs to read.

**Three comments describe an earlier design.** The workflow header opens "No golden images are committed" and `PresenterScreenshotTest`'s doc says "no golden is ever committed" — while this PR commits six PNGs, and `AGENT.md` and `build.gradle.kts` both correctly say they are committed and explain why. The same doc also names `build/outputs/roborazzi` as where a local run writes. Reworded all three to describe what ships: images are committed for diff reviewability, CI overwrites them before comparing, nothing reads the committed copy as a baseline. That distinction is the whole design, so it seemed worth having the comments agree on it.

**What I verified**
- Plain `./gradlew :composeApp:jvmTest --tests '*PresenterScreenshotTest*'` passes on macOS and does **not** touch the committed PNGs — so `:composeApp:check` is not broken for developers on a non-Linux machine, which was my main worry given AGENT.md's rule about pixel assertions.
- `recordRoborazziJvm --rerun-tasks` on macOS reproduced all six committed images **byte-identically**, so the goldens in git are consistent with what this toolchain produces.
- Merged current `main` (#221–#224 have landed since this branched) and ran the full suite: green.
- YAML parses.

One thing worth a second opinion rather than a change: the record step runs `recordRoborazziJvm --tests '*PresenterScreenshotTest*'`, which will rewrite the working tree's PNGs in CI. That is harmless as things stand, but if a future job ever runs `git diff --exit-code` to catch uncommitted regeneration, these two behaviours will collide.

PR #227, opened 2026-08-07
Roboratzii v2

grunt2001c:


zitlem:
Reviewed. The approach is sound and it fits the design #225 established — I checked the thing that most needed checking and it holds.

**The images are actually regenerated by CI.** I ran the workflow's exact filter (`recordRoborazziJvm --tests '*ScreenshotTest*'`) on this branch and both new classes are picked up:

```
PicturesTabScreenshotTest
PresentationTabScreenshotTest
```

So the comparison stays like-for-like on one runner, which is what makes the whole scheme work. Worth stating explicitly because it is a naming coincidence, not something enforced — see the note I left on #228, where it does not hold.

**The shared test-support changes are safe.** `width` and `themeMode` both default to null and preserve the old `MaterialTheme` path, and the full `test` check is green, so existing tests are unaffected.

**Cost is fine.** 38 tests in 9.2s; slowest single case 1.115s (`a folder deep enough to fill the grid`), which reads as first-test Compose warm-up rather than the test itself.

---

**One finding worth acting on: `RecentPictureFolders` is saved and restored by hand.**

`PicturesTabScreenshotTest` clears `RecentPictureFolders.folders`/`.pinned` in `@BeforeTest` and puts them back in `@AfterTest`. That is the exact pattern AGENT.md calls out — the objection is not that it is wrong today but that the restore is something each author has to remember rather than something structural.

`RecentFilesSwap` already exists for this and handles both lists *and* both file paths in one place:

```kotlin
RecentFilesSwap(
    readPaths = { RecentPictureFolders.file to RecentPictureFolders.pinnedFile },
    writePaths = { f, p -> RecentPictureFolders.file = f; RecentPictureFolders.pinnedFile = p },
    entries = RecentPictureFolders.folders,
    pinned = RecentPictureFolders.pinned,
    prefix = "cp-recent-picture-folders",
)
```

To be fair: I checked and **nothing here writes to disk** — `folders` is a plain `mutableStateListOf` and only `add`/`togglePin`/`clear` call `save()`, none of which this test calls. So the developer's real `~/.churchpresenter` is not at risk as written. The concern is that it is one careless edit away from being at risk, and that the helper for avoiding that already exists.

**Related, and needs fixing either way:** `PicturesTabTestSupport.kt` carries a doc comment stating it *"Deliberately never touches `RecentPictureFolders`"* and explaining at length why the Select Folder button and recent-folder chips are left untested for that reason. That comment is now false for tests built on this helper. Whichever way you go, it should be updated — a comment that documents a constraint the code no longer honours is worse than no comment.

**Minor:** `ThemedForTest` is duplicated verbatim in `PicturesTabTestSupport.kt` and `PresentationTabTestSupport.kt`.

**Worth being explicit about:** these tests assert nothing — they render and capture, by design. That is fine for their purpose, but they will move the JaCoCo number without adding verification, since a composable counts as covered on render alone. Worth keeping in mind when reading coverage after this lands.

**Not this PR's fault:** the `screenshots` check failure is infrastructure — details on #228, and it needs fixing on `main`.

Happy to push the `RecentFilesSwap` swap and the comment correction if you want them.

PR #228, opened 2026-08-07
Add app-preview screenshot tests and assets

grunt2001c:
Add a full app-preview screenshot test suite and supporting assets. New jvmTest cases (AppPreview*Tests) drive the UI to capture light/dark screenshots for multiple tabs (Bible, Songs, Media, Pictures, Presentation, Lower Third, Announcements, Timers) and a large AppPreviewSupport helper that seeds a sample library (songs, KJV bible, pictures, deck PDF, lower-third JSONs, welcome-loop.mp4) and captures images into composeApp/screenshots/app. Purpose: record deterministic Roborazzi/preview screenshots and provide preview resources for CI and review.


zitlem:
Reviewed. Two substantive problems, one of which means this PR's images cannot do the job they exist for.

## 1. None of these 63 images are ever regenerated by CI

`screenshots.yml` records with `--tests '*ScreenshotTest*'`. The classes here are named `AppPreviewMediaTest`, `AppPreviewSettingsTest`, and so on — **`AppPreview…Test` does not contain the substring `ScreenshotTest`**, so the filter never selects them.

Verified by running CI's exact command on this branch. Six classes ran; not one of this PR's:

```
BibleTabScreenshotTest
DropdownSelectorScreenshotTest
PresenterScreenshotTest
ScheduleTabScreenshotTest
SongsTabScreenshotTest
TabControlsScreenshotTest
```

The consequence is that `previewApp/` and `output/` are compared as *committed copies on both sides*. They change only when a human re-records locally, so the check can never surface a regression on those surfaces — the images are decorative rather than load-bearing. It also breaks the guarantee #225 was built around ("CI overwrites them before comparing, so nothing depends on which platform recorded the copy"), which now holds only for the classes that happen to match the filter.

Fix is either renaming the classes to `AppPreview…ScreenshotTest` or widening the filter. Renaming is probably safer — the filter is load-bearing and currently invisible.

For scale: re-recording #227's 16 pictures images on my macOS machine changed **15 of 16** files. Cross-platform drift on these is total, not marginal, which is exactly why the regeneration step matters.

## 2. The media preview test sleeps for 3 seconds

In `AppPreviewMediaTest`:

```kotlin
repeat(60) { Snapshot.sendApplyNotifications(); waitForIdle(); Thread.sleep(50) }
```

AGENT.md: *"**NEVER** use `Thread.sleep` or `delay` to wait for async work. A fixed pause asserts on timing, not behaviour, and flakes on a loaded CI machine. Wait for the condition itself."* There is no positive signal here — it waits a fixed 3s and hopes the video has rendered, so on a loaded runner it captures whatever happens to be on screen.

And it cannot succeed on CI anyway: **no workflow installs VLC**, so the video cannot play there at all. On every full-suite CI run this is 3 seconds spent waiting for something that cannot happen — and since these classes are not in the record filter, no capture is taken either.

This is the one test I would ask to be changed before merge, regardless of what you decide about the rest.

## 3. Cost

Against AGENT.md's ~1s-per-test bar:

| test | time |
|---|---|
| `the media tab` | **7.23s** |
| `the announcements tab` | **3.86s** |
| `the dictionary tab` | 1.11s |

The media one is mostly the sleep. The others are genuine composition cost — a different thing in kind, since the rule targets tests whose cost is a *duration* rather than work — but they are still a standing tax on a suite that runs constantly.

## 4. Repository growth

This takes `composeApp/screenshots/` from **3.6 MB to 12.5 MB** (75 → 138 files). Since git keeps every version of a binary forever, and a UI change rewrites many images at once, this compounds. Worth a deliberate decision now rather than discovering it at 200 MB. (The `welcome-loop.mp4` is only 24 KB — not a concern.)

## 5. Same caveat as #227

These assert nothing by design. Fine for their purpose, but they move the JaCoCo number without adding verification, since a composable counts as covered on render alone.

---

## Not this PR: the screenshots check is broken on main

Both PRs show a red `screenshots` check, and neither is at fault:

```
Error: Command 'git' failed with args 'fetch -u origin reg_actions:reg_actions':
       fatal: couldn't find remote ref reg_actions
fatal: unable to access '...': The requested URL returned error: 403
```

reg-actions v3 stores its baseline on a `reg_actions` **branch**, and the workflow grants `contents: read`, so it cannot create or push it — `git ls-remote` confirms that branch has never existed. The comment in the workflow says it *"stores the baseline images as workflow artifacts"*, which is what led to the wrong permission. It needs `contents: write`.

Until that is fixed the check is permanently red on every PR. It is advisory so it blocks nothing, but a check that is always red trains people to stop reading checks. I can push that fix to `main` separately if you want it.

zitlem:
Pushed fixes for both findings, plus one I hit while fixing them. Rebased onto `main` (#229 has landed, so images are no longer committed — that removes the repo-growth point entirely).

## 1. Naming — the classes now reach CI

Renamed all 15 to `AppPreview…ScreenshotTest`. Verified with CI's exact command: **59 preview images are now produced where there were none.** Before the rename the filter selected six classes and not one of these.

## 2. Paths — four literals were writing outside the compared root

`AppPreviewSupport`, `AppPreviewOutputTest`, `AppPreviewSettingsTest` and `AppPreviewRemotePermissionTest` each carried their own `"screenshots/…"` literal. After #229 that writes outside `SCREENSHOT_ROOT`, and an image there is not a *differing* image — it is one with no counterpart, which silently stops being compared. All four now use the shared constant.

## 3. The media test — removing the sleep crashes the JVM

This is the part worth reading, because my review of it was incomplete.

I said the `Thread.sleep` was pointless because CI has no VLC. That was true but not the whole story. Removing it **crashes the JVM**:

```
SIGBUS (0xa) … libvlccore.9.dylib vlc_object_hold+0x8
Process 'Gradle Test Executor' finished with non-zero exit value 134 (SIGABRT)
```

The double-click starts libvlc, and tearing the composition down while it is still initialising dies in native code. **The sleep was load-bearing** — keeping the process alive long enough for VLC to settle, not waiting for a frame. Confirmed by control: the original passes, the sleepless version crashes, so this was caused by the change and not a pre-existing flake.

So the fix is not a shorter wait, it is not starting playback at all. Beyond the crash, a decoded frame is the wrong thing to photograph: which frame arrives depends on machine speed, so every run would differ and the comparison would flag a change on every PR — and since CI can never reach that state, the baseline would only ever be reproducible on a developer's machine. The idle tab is identical everywhere, which is what a baseline has to be.

Runs 3× clean with no `hs_err` log. Full suite green.

## Incidental: four orphaned images

`timer_clock_*` and `timer_time_till_*` were committed but no test generates them — `AppPreviewTimersScreenshotTest` only defines `countdown` and `count_up`. They were leftovers from an earlier version of the test, sitting in git with nothing producing them. They simply vanish now that images are generated rather than stored, which is a fair argument for the change on its own.

## Still open, not addressed here

The cost of `the announcements tab` (3.9s) and `the dictionary tab` (1.1s) against AGENT.md's ~1s bar. Unlike the media one these are genuine composition cost rather than a duration, so I left them alone — but they are a standing tax on a suite that runs constantly, and worth a look if you think the states justify it.

zitlem:
Fixed the `screenshots` failure.

```
PicturesTabScreenshotTest > the animation picker FAILED
ComposeTimeoutException: Condition (no thumbnail still loading) still not satisfied after 5000 ms
147 tests completed, 1 failed
```

**This one is on me, and the cause is worth being precise about.** The test belongs to #227 and is already on `main`, where its own push runs pass. What changed is load: renaming the app-preview classes into the `*ScreenshotTest*` filter took the record job from ~6 classes to 21, and 147 tests on a two-core runner. The 5s wait was always marginal; the rename is what made it bite.

Traced rather than assumed — the earlier `app-previews` failures were the reg-actions 403, and the first run with a *test* failure is the first run after the rename.

Nothing about the test is wrong. It polls a positive condition and calls `Snapshot.sendApplyNotifications()` so the decode on `Dispatchers.IO` becomes visible. Only the threshold was too tight — and it had drifted: the same decode was allowed **10s** in `AppPreviewPicturesScreenshotTest`, and a deck rasterise **20s** in `AppPreviewPresentationScreenshotTest`. Identical work, three different numbers.

All five render waits now share `RENDER_TIMEOUT_MS`, so they cannot drift apart again.

**On AGENT.md's "never fix a flake by widening a timeout"** — I think this is the exception rather than a breach, and the reasoning is written at the constant so the next person can disagree with it knowingly:

- These waits end on their **condition**, so the timeout is only ever reached on the failure path. Measured after the change, the affected classes cost the same as before — `PicturesTabScreenshotTest` 2.7s, `PresentationTabScreenshotTest` 7.3s. A fast machine pays nothing.
- The wait is load-bearing for **correctness**, not just for passing. A thumbnail still showing its placeholder is a *different image*, so cutting the wait short would produce a screenshot that differs from run to run — the opposite of what a baseline needs.

Full suite green, 172 images recorded.

PR #229, opened 2026-08-07
Generate screenshots in CI instead of committing them

zitlem:
Drops the 75 committed PNGs and has CI generate them instead, and fixes the permission that has kept the `screenshots` check red on every PR since it landed.

## Why

Nothing in CI ever read the committed images as a baseline. The workflow records **both** the base branch and the branch on one runner, and reg-actions keeps its own baseline, so the copies in git only ever served the PR diff — which shows two opaque PNG blobs a reviewer cannot compare by eye, where the reg-actions comment puts before and after side by side.

Against that, they cost real churn. Skia rasterises text per platform, so re-recording the same *unchanged* states on another machine rewrote **15 of 16** files in one measured case, and git keeps every version of a binary for ever. The directory had reached 3.6 MB across 75 files, and a single UI change rewrites many at once.

## What changed

- Images go to `composeApp/build/screenshots`, defined once as `SCREENSHOT_ROOT`.
- `PresenterScreenshotTest` had its own literal copy of the old root; it now shares that constant. This matters more than it looks: images are matched between the two sides **by their path relative to the root**, so a capture written elsewhere is not reported as a differing image — it is one with no counterpart, and it quietly stops being compared.
- `permissions: contents: write`. reg-actions stores its baseline on a `reg_actions` **branch** it pushes to, not as a workflow artifact as the comment claimed. With `contents: read` every run failed with `couldn't find remote ref reg_actions` and then a 403; `git ls-remote` confirms that branch has never existed.
- AGENT.md updated: screenshots are not committed, captures must go under `SCREENSHOT_ROOT`, and a class must be named `…ScreenshotTest` or the workflow's `--tests '*ScreenshotTest*'` filter never renders it in CI.

## Verification

- Recording after the change produces **75 images whose relative paths are identical to the 75 removed** — diffed the two lists, no difference. So reg-actions still pairs base↔branch correctly rather than reporting everything as new.
- The old `composeApp/screenshots/` is not recreated by a record run, and is now gitignored so it cannot drift back in.
- Full suite green. YAML parses.

## Note for the open PRs

This removes the repo-growth objection from #228 entirely. Its naming problem still stands on its own — `AppPreview…Test` does not match `*ScreenshotTest*`, so those classes are never rendered in CI and their images are never compared, which is now the only thing that would make them real. #227 is unaffected: both its classes already match the filter.

One caveat: the baseline only exists once a push-to-main run has stored it, so the first PR run after this merges has nothing to compare against.


zitlem:
Rebased onto `main` now that #227 has landed, and dropped the 38 images it added — same reasoning as the rest of this PR.

Re-verified the property that matters with those included: recording after the change produces **113 images whose relative paths are identical to the 113 now on `main`**. Diffed the two lists, no difference, so reg-actions still pairs base↔branch rather than reporting everything as new.

Also worth recording: the permission fix is confirmed working, not just reasoned about — the previous run created the `reg_actions` branch (`git ls-remote` shows it), and the `screenshots` check went green for the first time. The no-baseline caveat I flagged did not bite, because that run established the baseline itself.

#227's two classes are named `PicturesTabScreenshotTest`/`PresentationTabScreenshotTest`, so they match the workflow's `*ScreenshotTest*` filter and are rendered in CI. Full suite green.

PR #241, opened 2026-08-08
Stop the coverage floor re-running the whole suite when the tests fail

zitlem:
## What happens today

`jacocoTestCoverageVerification` declares `dependsOn("jvmTest")`. That is right for a developer running the task directly. In CI the suite has already run in the step above, so:

- **suite green** → `jvmTest` is up to date, the dependency costs nothing, the step takes seconds. This is why the wiring looked fine for as long as the build was green.
- **suite red** → `jvmTest` is *not* up to date, so the Coverage floor step **starts the entire 21-minute suite again inside its 10-minute budget**. It cannot finish. Ever.

Observed on run [31232339922](https://github.com/ChurchPresenter/ChurchPresenter/actions/runs/31232339922) (`main`, 2026-08-08) — from the log, not from reasoning:

```
01:40:42  8844 tests completed, 1 failed, 4 skipped
01:40:44  > Task :composeApp:jvmTest FAILED
01:41:20  > Task :composeApp:jvmTest          <- the Coverage floor step, starting it again
01:50:57  ##[error]The action 'Coverage floor' has timed out after 10 minutes
```

So every red run pays **an extra ten minutes** and ends with a second failure reading *"Coverage floor timed out after 10 minutes"*, which carries no information about coverage. Whoever reads that log has to work out that the real failure was one test, twelve minutes earlier.

## The change

`-x jvmTest`, so the step can never re-run the suite. The verification reads `jvmTest.exec`, which the step above has already written.

Measured locally against existing exec data: **2 seconds**, and `:composeApp:jvmTest` does not appear in the executed task list.

It is also **advisory when the suite failed** (`continue-on-error` keyed off the test step's outcome). With a run that did not finish, the exec data is incomplete, so coverage is understated and a floor breach would be an artefact of the failure rather than a regression — and the build is already red for a better reason. When the suite passes, this gates exactly as before, and a coverage regression still fails the build.

The step keeps `if: always()`, so a genuine coverage regression is still reported alongside test failures rather than hidden behind them — which was the original intent, and is now actually achievable.

## Scope

`jacocoTestReport` has the same `dependsOn("jvmTest")` but is not invoked by this workflow, and no other workflow runs a task that depends on `jvmTest`. Checked rather than assumed. The Gradle task keeps its `dependsOn` so running it locally still does the right thing.

## What is NOT proven yet

CI on this PR exercises the **green** path only — which is precisely the path that always looked fine. The red path is what the change is for, so I am running a separate throwaway PR with a deliberately failing test to confirm from its log that the Coverage floor step does not relaunch the suite. I will post the result here before merging.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -66,6 +66,7 @@ jobs:
         run: chmod +x gradlew
 
       - name: App unit tests
+        id: unit_tests
         # Tighter than the job's own deadline so a hang here is attributed to this step and the
         # submodule suites below (`if: always()`) and the report upload still get to run. Runs land
         # around 11-14 minutes; one run was killed at a 25-minute budget with no diagnosis that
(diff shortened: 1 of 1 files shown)


zitlem:
## Red path confirmed — the path this change is actually for

CI on this PR only exercises the green path, which is the path that always looked fine. So I ran a throwaway PR (#242) off this branch with one deliberately failing test. Its `test` job log:

```
05:06:27  DeliberateFailureProbeTest > fails on purpose ... FAILED
05:23:15  > Task :composeApp:jvmTest FAILED
05:23:15  BUILD FAILED in 20m 28s
05:23:16  Run ./gradlew :composeApp:jacocoTestCoverageVerification -x jvmTest
05:23:21  > Task :composeApp:jacocoTestCoverageVerification
05:23:21  BUILD SUCCESSFUL in 5s
```

Step outcomes: `App unit tests` **failure** (20m28s) → `Coverage floor` **success, 5 seconds**.

`:composeApp:jvmTest` never appears in the Coverage floor step. Compare the same two steps on run 31232339922 before the change: the step relaunched the suite and was killed at its 10-minute budget.

| | before | after |
|---|---|---|
| Coverage floor on a red run | timed out after 10 min | **success in 5s** |
| Suite re-run by that step | yes, all 21 minutes | **no** |
| What the log said about coverage | nothing | the actual verification result |

Worth noting the verification **passed** on the partial exec data here, so `continue-on-error` did not have to do anything. It stays as insurance for a run that dies earlier and leaves thinner data, where a floor breach would be an artefact rather than a regression.

#242 is closed and its branch deleted.

PR #244, opened 2026-08-08
Fix three screenshots the website exports

zitlem:
Three captures in `previewApp/` are exported to churchpresenter.org, and all three were
publishing something the app does not actually look like. Found while re-importing the
site's screenshots.

### The schedule panel had no export shot at all

The website was using a hand-cut half of the stacked `scheduleTab/density_detailed.png`.
That fixture renders the panel at the full 1024dp test window against the 360dp an operator
really gives it (`AppPreviewSupport.library()`), so every row carried its text against the
left edge with most of the row empty — it read as a stretched, broken screenshot on the
homepage.

Worse, cutting it by hand meant nobody noticed it had been cut from a render taken *before*
#237 removed the fixture's placeholder paths. `/decks/sermon.pptx` and
`https://example.org/notices` were live on the marketing site while the app itself had been
fixed for a release.

So: one shot at 400dp, written straight into `previewApp/` beside the app-preview captures
it sits next to on the page, and written **unstacked** — the site needs the two themes as
separate files, and hand-cropping a stacked pair is what let a stale image through.

`scheduleTab()` grows an optional `density` for it, since `runComposeUiTest` has no density
hook and the site wants the same dp at more pixels. Null keeps the existing runner, so no
existing capture changes and reg-actions has nothing new to diff.

### The CCLI report was too narrow for its own table

Rendered at 940dp. The summary panel is a fixed `width(300.dp)` and the table takes
`weight(1f)` of the rest, but inside the table only title, author and songbook flex — rank,
CCLI number, count and the two dates are fixed, ~386dp once gaps and row padding are counted.
That left the three flexible columns 253dp between them:

| | at 940dp | at 1200dp |
|---|---|---|
| title | ~112dp — `Come Thou Foun…` | ~228dp |
| author | ~84dp — `Robert Robi…` | ~171dp |
| songbook | ~56dp — header read `SONGBO` | ~114dp |

Widening rather than shrinking, which is the intuitive move for a cramped screenshot and the
wrong one here.

The gutter still visible on the right is the row's own `end = 20.dp` scrollbar reserve plus
date columns wider than the dates — the dialog's layout at any window size, not something the
capture can fix. Left alone deliberately.

### The permission prompt carried a painted margin

`.background(surfaceVariant).padding(24.dp)` outside `.size(500.dp, 290.dp)`, so the PNG came
out 548x338 around a 500x290 dialog. Invisible against the dark site and an obvious grey slab
against the light one, fighting the frame the page already draws around it. Sitting flush is
safe because the content is an opaque `Surface` with no `shape`.

---

Test-only; no production code touched. Expect a visual difference on the CCLI and permission
shots — that is reg-actions working, not a regression.

### DIFF
--- a/composeApp/src/jvmTest/kotlin/org/churchpresenter/app/churchpresenter/screenshot/AppPreviewRemotePermissionScreenshotTest.kt
+++ b/composeApp/src/jvmTest/kotlin/org/churchpresenter/app/churchpresenter/screenshot/AppPreviewRemotePermissionScreenshotTest.kt
@@ -2,9 +2,7 @@
 
 package org.churchpresenter.app.churchpresenter.screenshot
 
-import androidx.compose.foundation.background
 import androidx.compose.foundation.layout.Box
-import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.size
 import androidx.compose.material.icons.Icons
 import androidx.compose.material.icons.filled.CalendarMonth
@@ -32,12 +30,16 @@ class AppPreviewRemotePermissionScreenshotTest {
             setContent {
                 ChurchPresenterTheme(themeMode = mode) {
                     Box(
-                        Modifier
-                            .background(MaterialTheme.colorScheme.surfaceVariant)
-                            .padding(24.dp)
-                            // The size the real DialogWindow opens at (DialogSizes.kt); without a
-                            // height the content stretches and leaves a gap the operator never sees.
-                            .size(500.dp, 290.dp)
+                        // The size the real DialogWindow opens at (DialogSizes.kt); without a
+                        // height the content stretches and leaves a gap the operator never sees.
+                        //
+                        // No backdrop behind it, and no padding around it: this shot is exported to
+                        // the website, which frames it itself. A painted margin inside the image
+                        // fought that frame — 24dp of surfaceVariant on all four sides, invisible
+                        // against the dark page and an obvious grey slab against the light one.
+                        // Safe to sit flush because the content is an opaque Surface with no shape
+                        // (RemoteEventDialog.kt), so there are no rounded corners to expose.
+                        Modifier.size(500.dp, 290.dp)
                     ) {
                         RemoteEventDialogContent(
                             event = event,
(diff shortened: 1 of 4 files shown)


zitlem:
Reviewed by running rather than reading. **Every claim in the description checks out, and I found nothing to fix.**

### The CCLI width diagnosis is exactly right

Rendered it at both widths to confirm the cause rather than take the arithmetic on trust:

| | 940dp (before) | 1200dp (after) |
|---|---|---|
| songbook header | **`SONGBO`** | `SONGBOOK` |
| row 2 title | `Come Thou Foun…` | `Come Thou Fount Of Every Blessing` |
| row 2 author | `Robert Robi…` | `Robert Robinson` |

The right-hand gutter you describe is visible in both and is the row's own scrollbar reserve plus over-wide date columns, exactly as documented — not something the capture width could fix. Correct call to leave it.

### The schedule export

400dp renders as a panel someone would actually use — rows fill their width, no stretched dead space. It also confirms the stale-crop story: the shot carries the post-#237 fixture (`/Users/Shared/ChurchPresenter/…`, `https://churchpresenter.org/notices`), not the `example.org` placeholders that had been live on the site.

Written unstacked into `previewApp/`, two files, one per theme — and it passes `ScreenshotInvariantsTest`, so the export is under `SCREENSHOT_ROOT` and will actually be compared rather than silently orphaned.

### "No existing capture changes" — verified byte-for-byte

The claim most worth checking, since it is what keeps reg-actions quiet. Recorded `scheduleTab/` on `main` and on this branch and compared checksums:

```
main: 14 files, branch: 14 files
diff → no output
```

**All 14 identical.** The `density = null` default really does keep the existing runner.

### The density knob does what it says

`-PpreviewDensity=2` (the Gradle property — `-D` goes to the daemon, not the test JVM, and quietly does nothing):

| capture | default | at density 2 |
|---|---|---|
| `previewApp/schedule_light.png` | 400x768 | **800x1536** |
| `scheduleTab/narrow_panel.png` | 240x1538 | 240x1538 |
| `scheduleTab/every_item_type.png` | 1024x1538 | 1024x1538 |

Scales the export, leaves the regression shots alone.

### The permission prompt

500x290 with no painted margin. Sitting flush is safe as described — the content is an opaque `Surface` with no `shape`, so there are no rounded corners to expose.

### Other checks

- `Schedule` + `Screenshot` + `Statistics` suites, 3 consecutive runs with `--rerun-tasks`: green each time. The `scheduleTab()` signature change breaks none of its other callers.
- `MaterialTheme` is still used at line 48 of the permission test, so removing the `background`/`padding` imports left nothing dangling.

One note, not a blocker: the branch is behind `main` by #241, so it may need updating before the required check will let it merge.

PR #246, opened 2026-08-08
Moved Screenshots back

grunt2001c:


zitlem:
Reviewed by running. **The change is competently made — it works, and it is internally consistent.** But I measured the cost it trades against, and the number is much larger than the one in the docs it replaces. That is a call for @zitlem, not for me, so I am reporting rather than merging.

## What works

- `ScreenshotInvariantsTest` was **inverted rather than deleted** — the naming and root-derivation invariants are intact and the committed-vs-`build/` assertion now guards the new direction. That is the right way to change an invariant.
- `SCREENSHOT_ROOT`, `roborazzi.outputDir`, `.gitignore`, the workflow and `AGENT.md` all move together. No stale comment left behind, which is what went wrong the last two times this area changed.
- `recordRoborazziJvm --tests '*ScreenshotTest*' --rerun-tasks`: green, writes all 247 into `composeApp/screenshots`.
- Full suite `--rerun-tasks --no-build-cache`: **green, 8m23s**.

## The measurement

I re-recorded on this machine (macOS) and compared, then downloaded the `screenshots` artifact from this PR's own CI run (`31252980923`, ubuntu) and compared that:

| comparison | images differing |
|---|---|
| committed set vs a **macOS** re-record | **21 of 247** (8.5%) |
| committed set vs **CI's Linux renders** | **246 of 247** (99.6%) |

Only **one** committed image is byte-identical to what CI produces.

So the committed set is a macOS recording, and it agrees with almost nothing the Linux runner renders. Two consequences worth weighing:

1. **A reviewer opening the committed PNGs is approving different pixels than reg-actions compares.** The workflow comment says as much — "the committed copies are for humans, not for this job" — but the two views can drift arbitrarily far apart without anything noticing, and here they already have.
2. **Any contributor recording on a different OS produces a 247-file diff**, every time, for no visual change. Your workflow comment already says "Record on ONE platform per branch" — that instruction is now load-bearing for the whole repo, and git keeps every version of every binary permanently.

For context, `AGENT.md` on `main` cites "15 of 16, measured" for this churn. **That understates it**: measured across platforms here it is 246 of 247.

## Not an objection

Reviewability is a real goal and the current setup does not serve it well — `build/` is git-ignored and wiped, so nobody but the recorder can open an image. That problem is genuine and this PR is a coherent answer to it.

I am only pointing out that the cost is ~99% churn per cross-platform record, not ~8%, and that this location has now moved three times. If it is going to be committed, it is worth deciding *which platform is canonical* and enforcing it, so the committed set and CI's renders do not silently diverge — otherwise the fourth flip is only a matter of time.

Happy to merge as-is if that trade is the one you want.

PR #247, opened 2026-08-08
Add lower-third screenshot and test

grunt2001c:
Add a committed lower-third preview image and a test that reliably captures it. Updates AppPreviewOutputScreenshotTest to: add a `lower third on screen` test, pre-render a Lottie frame via LowerThirdOffscreenRenderer (with a pump loop waiting for a non-empty frame), convert pixels into a Skia Bitmap and Compose ImageBitmap, and pass it to LowerThirdPresenter. Also adjust the test window size to OUTPUT_WIDTH/HEIGHT constants and add timing/constants. The new screenshot is composeApp/screenshots/output/lower_third.png to match repository screenshot conventions.


zitlem:
Reviewed by running, on head `47c68598` (re-validated after the `main` merge landed mid-review). **Good change — no defects found.**

## What I checked

**The pump loop is the right shape.** It ends on a positive signal (a frame with a non-transparent pixel) and the cap only ever *throws*, never lets the test pass — which is the distinction AGENT.md draws. I instrumented it to see the real headroom:

```
PUMP_FRAMES_USED=1 of 60
```

The first render is already non-empty, so there is 60× margin. No CI flake risk from a two-core runner being slower here.

**The image is right.** The emptiness guard only requires *some* pixel above alpha 8, which a nearly-blank frame would satisfy, so I opened the PNG: a properly composed band — "Pastor John Rivers / Guest Speaker", gold accent bar, correctly placed on the 1920×1080 output. Not a near-miss that slipped through the check.

**Cost.** `lower third on screen` comes in under 2s, in line with the nine siblings that each render a full 1920×1080 output; class total 17.0s for 10 locally.

**Stability.** `*AppPreview*` plus `ScreenshotInvariantsTest`, 3 consecutive runs with `--rerun-tasks`: green each time. Committed `lower_third.png` is byte-identical to a local macOS re-record.

Passing a pre-rendered frame instead of the live Compottie painter is well justified in the doc comment, and it matches what a warm output actually draws.

## One data point on the newly-merged committed-screenshot regime

Not an objection to this PR — it is correct under the policy that just landed in #246. But this is the first PR after it, so it is worth recording what happened:

I downloaded this branch's own CI artifact (run `31254906011`, ubuntu) and compared:

| | result |
|---|---|
| committed `output/` images vs **CI's Linux renders** | **0 identical, 10 differing** |
| the new `lower_third.png` specifically | differs — 22,946 bytes committed vs 22,812 from CI |

So the committed image a reviewer opens is a macOS render, and CI renders a different one every time. That is exactly the divergence I measured on #246 (246 of 247 across the whole set), now visible on a single new file.

It does not affect the correctness of this change. It does mean the "record on ONE platform per branch" line in the workflow is load-bearing, and that deciding which platform is canonical is still worth doing — otherwise every contributor on a different OS re-records a diff nobody wants.

Happy for this to go in once CI is green.

PR #250, opened 2026-08-08
Update Bible preview screenshot to Genesis 1:1

grunt2001c:
Change AppPreviewOutputScreenshotTest sample to Genesis 1:1 (KJV) and update the expected screenshot image to match the new verse text "In the beginning God created the heaven and the earth." This aligns the test data and committed screenshot.


zitlem:
Reviewed by running, on head `b4e698cd`. **No defects — merging.**

- **The committed image actually matches the code.** Re-recorded locally and `git status` on `composeApp/screenshots/output` came back empty, so `bible.png` is a genuine render of the new fixture rather than a stale or hand-edited file. That is the failure this regime is most prone to, so it is the first thing I checked.
- **The verse text is right.** Opened the PNG: "In the beginning God created the heaven and the earth. / Genesis 1:1", correctly laid out. It is the real KJV 1769 wording — *heaven*, not *heavens*.
- **Nothing else depended on Psalm 23.** The other `shepherd` references live in `BibleTabTestSupport`'s `SpbFixture`, a separate fixture for the Bible tab suites, so this changes only the preview.
- `*AppPreview*` plus `ScreenshotInvariantsTest`, 3 consecutive runs with `--rerun-tasks`: green each time.

PR #253, opened 2026-08-08
Extract the Browser Source content from its render coroutine, and test it

zitlem:
## The largest uncovered block outside `main.kt`

`BrowserSourceVideoRenderer.start` held **209 lines of Compose** as the content lambda of an `ImageComposeScene` inside a coroutine. Nothing could reach them — rendering them meant standing up the whole frame pump. JaCoCo scored `start$1` at **204 missed, 0 covered**.

But it is ordinary Compose: the identify overlay, the stage monitor, and the mode dispatch to each presenter. Only the pump around it — `scene.render`, the dirty-rect diff, the PNG encode — genuinely needs the scene, and that stays exactly where it was.

So the content moves **verbatim** into an `internal @Composable BrowserSourceContent`, mirroring the `…Content` split the dialogs already use (`PlanningCenterImportDialogContent`, `RemoteEventDialogContent`, `CCLIReportContent`).

It takes **only value parameters**, deliberately: AGENT.md warns that JaCoCo scores every lambda as its own method, so a helper taking function parameters can score *worse* than the code it replaced. That warning is the reason this was worth measuring rather than assuming.

## Measured

| | before | after |
|---|---|---|
| `start$1` | 204 missed / 0 covered | **43 missed** — the frame pump alone |
| all `BrowserSourceVideoRenderer*` | 215 missed / 95 covered | **125 / 220** |
| **project total** | 86.20% | **86.36%** — +0.160pp, 93 lines |

## The tests

Nine, rendering each branch that can end up on a live stream: the identify overlay (and that it names **only** the output asked for), a live verse, a live lyric section, the stage monitor, the blank state, and the announcement / Q&A / dictionary branches.

The device-backed modes (MEDIA, PICTURES, PRESENTATION, CANVAS, WEBSITE) are left out — they need VLC, a decoded image, a rasterised deck or a browser, which no split fixes.

One test was **rewritten after it failed**: it asserted the stage monitor does *not* show the live verse. That was wrong about the product — the platform needs the same words the room is reading. It now turns on the real difference: only the monitor looks ahead to the next verse.

## A regression I caused, and how it was caught

At the output's real 1920×1080, these nine tests **measurably loaded the suite**: two full-suite runs each failed a *different* unrelated test — `AtemClientSocketTest`, then `LowerThirdTabTest` — both green in isolation. A control run on unmodified `main` was green, which is what proved it was mine rather than the machine.

The tests assert that text reaches the semantics tree, not how it is laid out, so the surface is now **960×540** — same aspect ratio, a quarter of the pixels. **Two consecutive full suites green afterwards.**

## Validation

- Full suite `--rerun-tasks --no-build-cache`: **green ×2** (and red ×2 before the surface fix, so the fix is the difference).
- Control run on unmodified `main`: green — the failures were mine, not the machine.
- Coverage measured before and after; the numbers above are from the report, not an estimate.


zitlem:
Self-review, by running rather than re-reading.

**The "verbatim move" claim is verified, not asserted.** Diffed the old content lambda against the new composable body:

```
diff <(old start body, de-indented) <(new BrowserSourceContent body)
→ only a leading blank line
```

Modulo the `BrowserSourceVideoRenderer.` qualifier the companion helpers now need, the 209 lines are byte-identical. No behaviour change is possible from the move itself.

**Both top-level branches are mutation-caught**, each by exactly one test:

| mutation | caught by |
|---|---|
| `isIdentifying = false` | `identify covers the output with its own number` |
| `isStageMonitor = false` | `a stage-monitor assignment looks ahead, the congregation view does not` |

**Note for the reviewer:** this is based on `31b1de1d` and `main` has since moved to `8eb9a20d` (#251), so it will need `update-branch` — and that costs a fresh ~24 min `test` run before it can merge.

grunt2001c (approved):

PR #276, opened 2026-08-12
Give every icon-only button a name

zitlem:
Stacked on #275 — review that one first; this PR's diff is against its branch.

#275 labelled the three search clear buttons. Auditing for the same defect turned up **13 more** icon-only buttons drawn with `contentDescription = null` and nothing else: assistive technology meets an unlabelled button, and a test has no name to address one by.

| Pattern | Sites | Label |
|---|---|---|
| Pin/unpin star in recents | `MediaTab`, `PicturesTab`, `PresentationTab` | new `recent_pin`/`recent_unpin` |
| Clear-field buttons | `WebTab` ×2 (URL, type buffer) | new `web_clear_url`/`web_clear_typed_text` |
| Nav arrows | `CrosswordTab` ×2, `TabStripArrows` | new, 4 strings |
| Show/hide API key | `StockMediaBrowserDialog` | new, 2 strings |
| Add to schedule (row) | `SongsTab` | existing `add_to_schedule` |
| Cancel / confirm inline edit | `EditSongDialog`, `CanvasTab` | existing `cancel`; new `canvas_rename_confirm` |
| Loop toggle | `PicturesTab` | existing `loop_on`/`loop_off` |

The **pin toggles are the worst three**: a filled star and an outline star are the only thing separating pinned from unpinned, so without a description the *state* is unreachable, not merely the name — and it is the same copy-pasted code in three tabs.

Eleven new strings in `values/strings.xml` only; no other locale touched. Roughly half the sites reuse strings that already existed.

### Three tests selected on the absence of a label

`CanvasTabSceneRenameTest`, `PicturesTabLoopTooltipTest` and `SongsTabSelectionTest` addressed their button as *"the only unlabelled button"* — a deliberate, documented selector that asserted its own uniqueness. Labelling the buttons broke all three, loudly and exactly as those tests were designed to fail.

The selector is what was wrong, not the label. Each now addresses its button by name, which is both what an assistive-technology user gets and the more stable choice. Doc comments that recorded these buttons as unreachable are updated rather than left to mislead.

`canvas_rename_confirm` is a **new** string rather than a reuse of the pencil's `canvas_rename_scene` ("Rename"), so the control that *opens* the rename editor and the one that *commits* it stay distinguishable.

### Two deliberate exceptions

**The Songs table's action-column header keeps `contentDescription = null`.** Its click *sorts*, so naming it "Add to Schedule" would name it after an action it does not perform. Giving a sortable header its proper accessible name is a semantics question for the whole table, not for this icon — left alone on purpose, with a comment saying so.

**"Add to Schedule" is now correctly the name of several controls** — the toolbar button, one per table row, one per favourites entry. `SongsTabSelectionTest` used `onAllNodes(...)[0]`, which the new row buttons displaced. The shared name is right and stays; the toolbar button gets a test tag so the test can say which control it means instead of relying on tree order.

### Verification

- `./gradlew compileKotlinJvm` — green.
- `./gradlew :composeApp:jvmTest` — full suite green (9,814 tests).
- `./gradlew :composeApp:verifyRoborazziJvm` — no tab image moved. A content description draws nothing.
- A scanner for both shapes of this defect (`IconButton`, and clickable `Box`/`Row`/`Column`/`Surface`/`Card` containers) now reports **0** across `jvmMain`. It was validated against the pre-#275 tree, where it finds exactly the three buttons that PR fixes.

The same 5 pre-existing screenshot reds as #275 remain, unchanged and unrelated: `previewApp/ccli_report_{,bible_,activity_}light` (date-filtered), `canvasTab/source_camera` (enumerates real cameras), `lottieGen/section_shape` (order-dependent).

### Not included

Custom wrappers that already take a name — `TooltipIconButton`, `ActionIconButton` — were excluded because they supply their own description. I did not separately audit whether every *caller* passes a sensible one.

### DIFF
--- a/composeApp/src/jvmMain/composeResources/values/strings.xml
+++ b/composeApp/src/jvmMain/composeResources/values/strings.xml
@@ -2255,4 +2255,18 @@
     <string name="planning_center_import_matched">✓ Matched</string>
     <string name="planning_center_import_button">Import Selected</string>
     <string name="planning_center_import_file_count">%1$d file(s)</string>
+
+    <!-- Names for buttons whose only content is an icon. These are read by assistive technology
+         and are never drawn, so a button that already shows a text label does not need one. -->
+    <string name="web_clear_url">Clear address</string>
+    <string name="web_clear_typed_text">Clear text</string>
+    <string name="recent_pin">Pin</string>
+    <string name="recent_unpin">Unpin</string>
+    <string name="canvas_rename_confirm">Confirm rename</string>
+    <string name="crossword_prev_level">Previous level</string>
+    <string name="crossword_next_level">Next level</string>
+    <string name="tab_strip_scroll_back">Scroll tabs back</string>
+    <string name="tab_strip_scroll_forward">Scroll tabs forward</string>
+    <string name="stock_photo_show_key">Show API key</string>
+    <string name="stock_photo_hide_key">Hide API key</string>
 </resources>
--- a/composeApp/src/jvmMain/composeResources/values/strings.xml
+++ b/composeApp/src/jvmMain/composeResources/values/strings.xml
@@ -2263,6 +2263,7 @@
     <string name="recent_pin">Pin</string>
     <string name="recent_unpin">Unpin</string>
     <string name="canvas_rename_confirm">Confirm rename</string>
+    <string name="qa_clear_question_text">Clear question text</string>
     <string name="crossword_prev_level">Previous level</string>
     <string name="crossword_next_level">Next level</string>
     <string name="tab_strip_scroll_back">Scroll tabs back</string>
--- a/composeApp/src/jvmMain/kotlin/org/churchpresenter/app/churchpresenter/composables/TabStripArrows.kt
+++ b/composeApp/src/jvmMain/kotlin/org/churchpresenter/app/churchpresenter/composables/TabStripArrows.kt
@@ -14,10 +14,14 @@ import androidx.compose.ui.unit.dp
 import churchpresenter.composeapp.generated.resources.Res
 import churchpresenter.composeapp.generated.resources.ic_arrow_left
 import churchpresenter.composeapp.generated.resources.ic_arrow_right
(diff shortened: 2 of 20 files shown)


zitlem:
Pushed `f54738cb`, closing the coverage gap noted at the bottom of the description.

**Both halves of the gap, checked.**

*Wrappers that take a name as a parameter* — `TooltipIconButton`, `ActionIconButton`, `GoLiveButton`, `AddToScheduleButton`, `PanelResizeHandle`: **92 call sites, all clean.** Every one of those parameters is required, so no caller can omit it, and every caller passes a real localised string. No change needed. (`RemoteEventDialog` and `QATab` each define their own local `ActionIconButton`/`QAIconButton`; both label their icons too.)

*Wrappers that take the icon as a content slot* — this was the actual blind spot. The original scan matched a literal `IconButton(`, so a button built from `QAIconButton { Icon(...) }` or `Tooltip { FilledIconButton { Icon(...) } }` was invisible to it. Scanning instead for **any call carrying an `onClick` whose icons are all unlabelled** finds them regardless of the callee’s name, and turned up 5 more:

| Site | Control | Label |
|---|---|---|
| `RemoteActivityToast.kt:270` | block-for-session | existing `block_for_session` |
| `LowerThird.kt:916` | ATEM go-live key toggle | existing `atem_golive_key` |
| `LowerThird.kt:938` | ATEM quick still upload | existing, hoisted |
| `LowerThird.kt:943` | ATEM quick clip upload | existing, hoisted |
| `QATab.kt:428` | clear add-question field | new `qa_clear_question_text` |

The three ATEM buttons already computed a tooltip string — in one case a four-branch `when` covering unreachable/clip-too-long/normal. Rather than duplicate that expression into the description, each is hoisted into a `val` used for both, so the tooltip and the accessible name cannot drift apart.

Both scans now report **0** across `jvmMain`. Two new strings total (`qa_clear_question_text`, plus `canvas_rename_confirm` from the earlier commit), English only.

Full suite green; screenshot verify shows the same 5 pre-existing reds and no new ones.

PR #357, opened 2026-08-21
Rename presentation engine package

grunt2001c:
The presentation-engine module was moved from the old `presentation.engine` package to `org.churchpresenter.presentationengine`, and the app, tests, and Gradle entry points were updated to match. This aligns the module with the repository's package conventions and avoids collisions with app-level `presentation` variables and identifiers.

PR #358, opened 2026-08-21
Stop two ATEM test suites paying a timeout in teardown

zitlem:
Two suites in `jvmTestSerial` ended tests by waiting out a timeout rather than on a signal — what `AGENT.md` rules out in as many words: *"It must never end by the timeout expiring; the timeout exists only to fail the test."* Between them they cost **33 seconds of every run**.

Both were found by measuring, not by reading: `./gradlew :composeApp:jvmTest` then the `time=` attributes in `composeApp/build/test-results/`. The giveaway in each case was a suspiciously round per-test number.

## `LowerThirdSequencerKeyTest` — 24.04s → 0.15s

Three tests took exactly 8.00s each; the other five took ~0.00s. The split is exactly whether a test leaves its sequence running.

The tests wrap the fake switcher in `use {}`, so it closes when the test body ends. But the sequence is still on air, and `@AfterTest` then calls `LowerThirdSequencer.stop()`. `stop()` drives a key-off at the switcher and *awaits its ack* — deliberately, because preemption is only deterministic if the previous key is off before the next goes on. A switcher that has already been closed acks nothing, so that wait ran out `AtemClient.CMD_TIMEOUT_MS` (8s) and `releaseKeyLocked` swallowed the resulting failure.

So teardown was exercising the timeout path rather than the key-off it exists to perform, and the three tests that left a sequence running paid 8s each.

A `withSwitcher` helper now stops the sequencer *inside* the `use` block, while the switcher is still answering, then lets it close.

## `AtemUploadTracedTest` — 11.7s → 1.7s

Five tests at a uniform ~2.3s. `Sentry.close()` blocks draining its queue for `shutdownTimeoutMillis` — default 2000ms — once per test. The transport is already `NoOpTransportFactory`, so there was never anything to drain; the wait was pure teardown cost. Set to 0.

## Why this is filed as a flake fix

`CompanionServerAtemUploadTest` has been failing on CI with `timed out after 5000ms waiting for N commands, got 0` — all five tests at once, the failure the `serialTestClasses` comment in `composeApp/build.gradle.kts` already documents. `AtemConnectionManager` is a singleton, and `@AfterTest` only invalidates it *after* `stop()` returns. While a teardown sat in that 8-second timeout, the shared client was in a failing state that the next suite could inherit.

I want to be straight that this is a plausible mechanism rather than a proven one: I could not reproduce the `got 0` failure locally, including under full CPU saturation. What is measured and certain is that both suites no longer wait on a dead socket, and that 33s of wall clock is gone.

## Not fixed here

A CI run on #357 hung for 22 minutes and took the 40-minute job timeout. The last logged test was `AppPreviewWebScreenshotTest`; in local ordering the class immediately after it is `LowerThirdTabScreenshotTest` — one of the three suites the build file already flags for opening the ATEM upload dialog behind three 5-second deadlines. I could not reproduce the hang (that suite exits cleanly here, and there are no non-daemon threads in `jvmMain`), so I have left it alone rather than guess at a fix. Recording the lead so the next occurrence starts further along.

## Verification

- Three consecutive `--rerun-tasks` runs of all five ATEM suites, green each time with stable timings
- The whole `:composeApp:jvmTestSerial` task passes
- `./gradlew :composeApp:detekt` clean

### DIFF
--- a/composeApp/src/jvmTest/kotlin/org/churchpresenter/app/churchpresenter/server/AtemUploadTracedTest.kt
+++ b/composeApp/src/jvmTest/kotlin/org/churchpresenter/app/churchpresenter/server/AtemUploadTracedTest.kt
@@ -23,6 +23,10 @@ class AtemUploadTracedTest {
             options.isEnableUncaughtExceptionHandler = false
             options.isEnableAutoSessionTracking = false
             options.tracesSampleRate = 1.0
+            // Sentry.close() blocks for this long draining its queue, and the default is 2000ms —
+            // paid by every test in this class, which is the whole of its runtime. There is nothing
+            // to drain: the transport above is NoOp, so the wait is pure teardown cost.
+            options.shutdownTimeoutMillis = 0
         }
     }
 
--- a/composeApp/src/jvmTest/kotlin/org/churchpresenter/app/churchpresenter/server/LowerThirdSequencerKeyTest.kt
+++ b/composeApp/src/jvmTest/kotlin/org/churchpresenter/app/churchpresenter/server/LowerThirdSequencerKeyTest.kt
@@ -30,6 +30,28 @@ class LowerThirdSequencerKeyTest {
         keyPostRollMs = 0,
     )
 
+    /**
+     * A fake switcher for one test, torn down in the order the sequencer requires: stopped while
+     * the switcher is still answering, and only then closed.
+     *
+     * [LowerThirdSequencer.stop] drives a key-off at the switcher and awaits its ack, deliberately
+     * — preemption is only deterministic if the previous key is off before the next goes on. A
+     * switcher that has already been closed acks nothing, so that wait runs out the client's full
+     * `CMD_TIMEOUT_MS` (8s) and `releaseKeyLocked` swallows the resulting failure. Closing the fake
+     * first therefore cost 8 seconds per test that left a sequence running — three of them here,
+     * 24 of this suite's 24 seconds — and bought a teardown that exercised the timeout path rather
+     * than the key-off it is meant to perform.
+     */
+    private fun withSwitcher(block: (FakeAtemSwitcher) -> Unit) {
+        FakeAtemSwitcher().use { fake ->
+            try {
+                block(fake)
+            } finally {
+                runBlocking { LowerThirdSequencer.stop() }
+            }
+        }
+    }
+
     private fun dskCommands(fake: FakeAtemSwitcher) =
         fake.commandsNamed("CDsL") + fake.commandsNamed("DDsA")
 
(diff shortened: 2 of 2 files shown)


zitlem:
CI hit the 40-minute timeout on this branch — the same pre-existing hang described under "Not fixed here", not a regression from this change.

The evidence that it is not this diff: the hang begins immediately after `AppPreviewWebScreenshotTest` completes, and **no ATEM suite reported at all** in that run. Both files this PR touches (`AtemUploadTracedTest`, `LowerThirdSequencerKeyTest`) are ATEM suites that had not yet executed when the JVM stopped making progress. The same hang, with the same signature and the same last-logged test, also occurred on #357 before this branch existed.

Two runs now share the signature exactly:

| run | last test logged | silence |
|---|---|---|
| #357 | `AppPreviewWebScreenshotTest` 13:01:22 | 22 min → 40-min job timeout |
| #358 | `AppPreviewWebScreenshotTest` 16:53:19 | 22 min → 40-min job timeout |

**New lead.** In local `jvmTestSerial` ordering the class immediately after `AppPreviewWebScreenshotTest` is `LowerThirdTabScreenshotTest`, and that suite's helpers call Compose's `waitForIdle()` — which, unlike the `waitUntil` calls around it, takes no timeout. A composition that never goes idle (the lower-third preview animates) would block there for ever and produce exactly this: no failure, no further output, silence until the job is killed. That is a hypothesis from ordering observed locally, not a confirmed diagnosis — Gradle's class ordering is not guaranteed to match on CI, and the alternative candidate by alphabetical order is `AtemUploadTracedTest`.

Re-running CI. If the hang recurs it deserves its own fix rather than another re-run.

zitlem:
Review note — a production issue this change surfaced but does not fix.

`stopLocked()` runs `releaseKeyLocked()` **before** `onClear.tryEmit(Unit)`:

```kotlin
generation++
job?.cancel(); job = null
releaseKeyLocked()          // awaits the switcher's ack
_status.value = "idle"
onClear.tryEmit(Unit)       // ← what actually clears the screen
```

When the ATEM is unreachable — unplugged, powered off, wrong IP — that ack never comes and the wait runs out `AtemClient.CMD_TIMEOUT_MS` (8s). Because `onClear` is emitted afterwards, **the lower third stays on screen for those 8 seconds**. `POST /api/lowerthirds/hide` also only responds once `stop()` returns, so a Companion/Stream Deck "hide" button reads as dead for 8 seconds — in precisely the situation where an operator will be pressing it repeatedly.

This is the same timeout the test change above works around, and the test suite was the only place it was visible.

Deliberately **not** fixed in this PR, which is test-only. Reordering the on-screen clear ahead of the switcher ack changes live output behaviour and deserves its own change: the question is whether the clear should simply precede the key-off, or race it against a much shorter deadline so a healthy switcher still gets a deterministic off-before-on for preemption. Filing the observation so it is not lost.

PR #360, opened 2026-08-21
Extract Bible formats into new Gradle module

grunt2001c:
Creates a new `:bible-formats` module and moves Bible catalog sources, install support, and `.spb` converters out of `:composeApp`/`:converter` into a shared library package. Updates app and converter dependencies/imports to consume the new module, adds module docs/build config, wires it into settings, CI, and `test-changed.sh`, and migrates/expands tests (cache behavior, CSV parsing, DTO construction, install helpers, path-language fallback, and rate-limit/not-modified index handling) while removing duplicate coverage from `composeApp`.


zitlem:
Reviewed. The extraction itself is clean — the moves are 78–99% renames, the `report_paths`/`check_name` lists are appended in the same order so the index pairing holds, and the `catch` narrowing that came with it keeps `UnresolvedAddressException` ahead of `IllegalArgumentException` everywhere, which is what stops an offline operator being told their download is malformed.

**One thing I fixed rather than just flagged, because it was a silent regression:** the new `bibleformats:` key was inserted directly above `planningcenter`'s three transitive filter lines, so those lines became bible-formats' and `planningcenter` was left watching only its own directory. They *are* correct for `:bible-formats` (it takes `:settings` and `:diagnostics` too) — but `:planning-center` takes the same two, and had them for the same reason, added earlier today in #343. Left as it stood, a change to `:settings`, `:diagnostics` or `:core-models` would silently stop running the Planning Center suite. Both modules now name all three. Pushed as one commit on this branch; revert it if you would rather do it yourself.

Pushing also re-triggered CI, which had been sitting at `action_required` since your detekt fix, so this is the first run that will actually exercise `:bible-formats:detekt` and its coverage floor.

Two notes, neither blocking:
- `bible-formats/build.gradle.kts` names no `coverageFloors`, so it takes the default 85% on all six counters. This run is the first check of that.
- The root `AGENT.md` sentence listing which modules name floors ("…`:settings`, `:diagnostics` and `:planning-center` name none") doesn't mention `:bible-formats` yet.
- `ZefaniaSource.kt`'s first catch block goes `IOException, SAXException, IllegalArgumentException` with no `UnresolvedAddressException`. If that block can be reached with the network involved, an offline user gets the malformed-response path; if it is parse-only, it is fine as is.

PR #361, opened 2026-08-21
Make a CI failure diagnosable: name classes at start, bound hangs, and fix two flakes

zitlem:
Four changes, all of them about CI telling the truth about itself.

## Two flakes, both root-caused and reproduced

**`CompanionServerPresentationRenderTest` — all four tests, `Connection refused`.** Failed twice today on CI, green on re-run and locally, and it is what blocked this PR's own last two runs.

`CompanionServer.start` does not bind the port you give it — it runs it through `findFreePort`, which walks upward when the port is taken (a previous suite's socket still in `TIME_WAIT` is enough). The server was up and `isRunning` was honestly true; it was just listening one port along from the constant every URL in the class was built from. Every sibling suite already reads the real port back out of `server.serverUrl`, and `CompanionServerTest` says why in a comment. This suite waited the same way and then threw the answer away.

Reproduced by holding all four per-fork ports (39890 / 40890 / 41890 / 42890) so `findFreePort` has to walk: **unfixed, all four tests fail exactly as on CI; fixed, they pass.** Three consecutive green runs after.

**`BibleViewModelChapterTest > a multi-selection formats as the range the audience sees` — `expected:<1-3> but was:<>`.**

`loadBibles` publishes a books-only module first and the verse text second, so `books.isNotEmpty()` is satisfied while chapters still hold no verses — the same window #344 fixed `resolveVerseSelection` for. A chapter opened in it publishes an empty verse list and bumps the selection token anyway, so `openChapter` returns satisfied, `shiftClickVerse` finds its target out of range and returns early, and the selection is empty. **Every other `BibleViewModel` suite already follows the books wait with a wait on `isFullyLoaded`; this was the only one that did not.**

## The hang: bounded and instrumented, still unexplained

Three runs have ended as a step timeout rather than a failure — 2026-07-30 (never explained), #344's screenshots job, #359's test job. The listener names each class **as it finishes**, which is exactly the event a hung class never reaches, so both August logs stop after the last class that *completed* and the hung one had to be inferred from execution order (`LowerThirdTabScreenshotTest`, both times).

- **A `START` line per class**, so the next occurrence ends with a `START` and no matching result.
- **A 30-minute task timeout**, so a hang fails in 30 with the log intact instead of running to the step's 40-minute limit and uploading no XML.

The timeout is a watchdog on the harness, not a wait inside a test. **It is 30 and not 20 because the `App unit tests` step now measures 18m45s and 19m24s on green runs** — the step's own comment still claims 11–14 minutes, that band is stale, and a 20-minute cap (tried first here) would have failed both of those healthy runs.

This does not explain the hang. The suite passes locally in 2.4s, still passes under a deliberately loaded machine (8.4s), and two probes of the frozen-clock states — a mouse move injected mid-tween, and a scene torn down with a tween and a pending 8s delay outstanding — both completed in milliseconds.

## Testing

Both fixed suites green three consecutive times with `--rerun-tasks`; `:composeApp:jvmTestSerial` green with a `START` and a result line for all 30 classes.


zitlem:
Retitled and amended: the cap is now **30 minutes**, not 20.

Measuring the `App unit tests` step on two recent green runs gives **18m45s** and **19m24s** (2026-08-21). The step's own comment still says runs "land around 11-14 minutes" — that band is stale, the suite has grown into it, and a 20-minute cap would have failed both of those healthy runs. 30 still ends a hang ten minutes before the step's 40-minute budget does.

Also worth noting for whoever reads this next: the failure on the previous run here was `BibleViewModelChapterTest > a multi-selection formats as the range the audience sees` (`expected:<1-3> but was:<>`), and `main` itself is currently red on four `CompanionServerPresentationRenderTest` tests. Both sets pass locally on `bbe40010`, so both are flakes rather than regressions — and neither is related to this change, which touches only the build file.

PR #413, opened 2026-08-26
Add NDI output

### LINKED ISSUE #373: NDI support
NDI (Network Device Interface) integration
Network Device Interface (NDI) to send clean lower-third lyrics to Streaming Softwares

zitlem:
Closes #373.

Sends this app's live content over the network as an NDI® source, so OBS, vMix or a hardware
switcher can take lyrics or a lower third straight off the LAN with no capture card.

**The mode that matters is alpha**, and it is the default: one sender carrying genuine per-pixel
transparency, so the clean lower third the issue asked for arrives in OBS already keyed — no second
source, no downstream keyer. SDI physically cannot do that, which is what makes this more than
"DeckLink without a card". Fill-only and a discrete fill + key pair are there for gear that expects
them.

## No NDI binaries are shipped, and none can be

Vizrt's licence does not permit redistributing the NDI Runtime. That constraint shapes the whole
design rather than being worked around: `libndi` is loaded at run time and the user is pointed at
the download when it is missing — the same resolution `obs-ndi`/DistroAV reached.

So **NDI behaves exactly as VLC already does here**: a separately installed runtime, auto-detected
at startup, with a path override in settings, and the feature reading as *not installed yet* rather
than as a failure when it is absent. Discovery honours `NDI_RUNTIME_DIR_V6`/`_V5`, so a machine
already set up for OBS's NDI plugin needs nothing done to it.

Two licence terms are implemented as terms, not as polish: the `NDI®` attribution line in the About
box, and the ndi.video link wherever NDI is offered.

## Shape

**1 · `ComposeScenePump` — extracted first, on its own.** There were three independent copies of
"render Compose off-screen, read the pixels, push them somewhere". Rather than add a fourth, the
generic half of `BrowserSourceVideoRenderer.start` — the scene, the virtual clock, the reused pixel
buffer, the idle parking — became `ComposeScenePump`, and Browser Source's dirty-rect/encode stage
became its `onFrame` callback. Its existing `shouldRenderTick`/`decideTick` tests are untouched and
still pass; the pump gained its own first coverage. `DeckLinkComposeOutput` is deliberately left
alone — its offscreen-`JFrame`+`SkiaLayer` shape exists to dodge a `GraphicsLayer.toImageBitmap()`
race and should not be propagated.

The content composable both renderers now share was renamed `BrowserSourceContent` →
`OffscreenOutputContent` and moved to its own file: it is no longer Browser-Source-specific, and the
old name already collided with a private composable of the same name in `SceneSourceRenderer.kt`.

**2 · `:ndi`** — plain `kotlin("jvm")` on the `:atem` recipe. Every native call sits behind one
`NdiLibrary` interface, so the suite drives the module against a plain Kotlin fake rather than a
mock. JNA over Devolay (unmaintained; no Apple Silicon native; hardcodes the `libndi.so.5` soname
NDI 6 dropped) and over Panama/FFM (JDK 22; this build targets 21). Its version moved into
`libs.versions.toml`, and `:composeApp` moved off its hand-copied literal onto the same alias.

**3 · Settings** — `ndiOutputs: List<ScreenAssignment>` plus an `ndi*` field tail, mirroring
`browserSourceOutputs`. Not `screenAssignments`: that list is reconciled against detected hardware
and drives the window-count arithmetic, and an NDI output maps to no display. Every field is
defaulted, so **no `CURRENT_SETTINGS_VERSION` bump and no migration** — there is a test pinning
that.

**4 · Wiring** — registered beside the Browser Source block in `main.kt`; `PresenterWindows.kt` is
untouched, since an NDI output opens no window. A JVM shutdown hook destroys the senders, as
`DeckLinkManager` does, so quitting removes the source instead of leaving a frozen frame on the
network. The runtime is brought up on `Dispatchers.IO`, not inline in the composition.

**5 · A card in the Projection tab**, not a new settings tab — `OptionsDialog`'s tab indices are
hardcoded in three places plus `companionSatelliteTabIndex` and the screenshot tests.

## Notes for review

- **Two bugs were found by writing the tests, not after.** `Memory(0)` throws rather than allocating
  nothing, so a misconfigured zero-dimension output took the render loop down; and the test fixture's
  plain lists raced the render coroutine. Both are fixed at source.
- **The NDI runtime state is a parameter, not a lookup, everywhere a screenshot can see it.** Left
  reading the live manager, the settings card would draw whatever the recording machine happens to
  have installed — the same nondeterminism that had `canvasTab/source_camera` enumerating real
  cameras. `ProjectionSettingsTab` takes it the way it already takes `detectScreens`.
- **The detekt baseline was edited by hand, never regenerated.** Six entries were re-keyed for code
  that moved file or changed signature; two were deleted because the lines no longer exist. Net
  −2 entries.
- `:ndi` meets the default 85% floor on **all six counters with no override and no exclude**. That
  is earned by covering the JNA marshalling against a fake `NdiLibC` rather than excluding it — and
  that exercise is what surfaced the `Memory(0)` crash an exclude would have hidden.
- The 21 modified `projectionSettingsTab` images changed only because the tab now has one more card.
  7 new images cover the card's states, in both themes.
- **The About window grew 490dp → 520dp.** `DialogViewportTest` measured its content at 504dp once
  the trademark line joined the copyright, i.e. 14dp off the bottom at the 1.3× text growth that
  suite treats as headroom. That test caught it; the fix is the window, not the test.
- Two "Browse" buttons on one tab would have been ambiguous to a user as well as to a test, so the
  NDI folder picker is labelled "Choose Folder". The path field itself is read-only with a chooser
  beside it — the shape the VLC path on this same tab already uses.

## Not in scope

**Audio**, deliberately: the app plays media through a single VLC player, and vlcj's audio callback
*replaces* the audio output module, so a naive PCM tap would silence the speakers. Solvable with a
dedicated audio-only player kept in sync — a second piece of work, not a footnote on this one.

**NDI receive**, deliberately: it belongs as a `SceneSource` in the Canvas compositor beside the
camera and screen-capture sources, and the cost is not the binding but a second live media pipeline
with its own reconnect and drop behaviour. The shape is recorded in `ndi/AGENT.md` so it is not
re-derived.

## Testing

Live sending needs both a runtime and a receiver, so it is not a unit test — verify by hand with
NDI Studio Monitor or OBS. Everything else is covered:

- `./gradlew :ndi:check` — 87 tests, no runtime and no network needed
- `./gradlew :settings:check`
- `./gradlew :composeApp:check`
- `./gradlew :composeApp:detekt` — clean
- New/changed `:composeApp` suites run three consecutive times for flakes, green each time


zitlem:
Superseded by #415.

This PR was branched off a local `main` that was 114 commits ahead of `origin/main` — including PRs GitHub records as merged (#405, #407) but which are not on the remote branch. GitHub's merge-base was therefore the stale `origin/main`, so the diff showed 3,224 files instead of the 84 this change actually touches.

#415 is the same feature ported onto `origin/main` and opened against it: 84 files, one commit, full suite green.

**The branch `ndi-output` is deliberately left undeleted.** It is currently the only copy on the remote of those 114 commits (7 modules — `web-tab`, `qa-tab`, `stt-tab`, `shortcuts`, `announcements-tab`, `ui-components`, `lower-third-tab`). Please don't prune it until it's clear whether that rollback of `main` was intended.

Issue #403, opened 2026-08-25
Song sections are reordered after saving instead of preserving the entered sequential order


phantomcodeX-code:
## Describe the bug

When creating or editing a song, ChurchPresenter does not preserve the sequential order of the song sections after saving.

The problem becomes noticeable when a song contains multiple verses and multiple chorus sections that are intentionally separated into different parts.

For example, the song may be structured as:

V1
- MUHATAG SA KALIPAY
- DIHA SA GINOO
- MUHATAG SA KALIPAY
- DIHA SA GINOO

C1
- DAYGON TA ANG DIYOS
- MUSALIG KANIYA
- ANG MALIPAYONG KASING KASING
- HALLELUJAH

V2
- MUHATAG SA KALIPAY
- DIHA SA GINOO
- MUHATAG SA KALIPAY
- DIHA SA GINOO

C2
- DAYGON TA ANG DIYOS
- MUSALIG KANIYA
- ANG MALIPAYONG KASING KASING
- HALLELUJAH

The intended purpose is to keep the song sections sequential and allow individual parts of the same Verse or Chorus to be separated into presentation slides.

For example, C1 may intentionally be divided into:

CHORUS 1 - SECTION 1
DAYGON TA ANG DIYOS
MUSALIG KANIYA

CHORUS 1 - SECTION 2
ANG MALIPAYONG KASING KASING
HALLELUJAH

These are still part of CHORUS 1. They are NOT separate song sections such as Bridge.

The same structure should be possible for C2, C3, etc.

---

## To Reproduce

1. Open ChurchPresenter.
2. Go to Songs.
3. Create a new song.
4. Add lyrics using Verse and Chorus sections.
5. Arrange the song intentionally in sequential order, for example:

V1
C1
V2
C2
V3
C3
Ending

6. Within a Verse or Chorus, intentionally separate the lyrics into smaller presentation sections/slides.
7. Save the song.
8. Open the saved song again.
9. Observe the generated song slides/sections.

---

## Expected behavior

The song should preserve the exact sequential order entered by the user.

For example:

V1
  Section 1
  Section 2

C1
  Section 1
  Section 2

V2
  Section 1
  Section 2

C2
  Section 1
  Section 2

V3
  Section 1
  Section 2

C3
  Section 1
  Section 2

Ending

A manual separator such as `[---]` should be treated as a slide/presentation break inside the current section, rather than changing the section itself.

For example:

[Chorus]
[...]

PR #448, opened 2026-08-30
Break a verse or chorus across slides without splitting the section

### LINKED ISSUE #404: Enhancement: Allow manual slide breaks within Verse and Chorus sections while preserving section sequence
## Feature request

I would like to suggest an enhancement to the Song Editor and presentation workflow.

The main purpose is to allow users to split a Verse or Chorus into multiple presentation slides while keeping all of those slides under the same song section.

For example, a Chorus may contain:

CHORUS 1

DAYGON TA ANG DIYOS
MUSALIG KANIYA

ANG MALIPAYONG KASING KASING
HALLELUJAH

Both parts are still CHORUS 1.

The user should be able to manually decide where the lyrics are divided into separate presentation slides without creating a new song section.

---

## Proposed syntax

A manual separator such as:

[---]

could be used to indicate a presentation slide break.

Example:

[Chorus]
DAYGON TA ANG DIYOS
MUSALIG KANIYA

[---]

ANG MALIPAYONG KASING KASING
HALLELUJAH

This should produce:

CHORUS 1 — SLIDE 1
DAYGON TA ANG DIYOS
MUSALIG KANIYA

CHORUS 1 — SLIDE 2
ANG MALIPAYONG KASING KASING
HALLELUJAH

The second part should NOT become a Bridge, Verse 2, or another Chorus section.

It should remain part of CHORUS 1.

---

## Sequential section numbering

It would also be useful to support repeated sections as separate sequential instances.

For example:

V1
C1

V2
C2

V3
C3

Ending

Where:

- V1 = Verse 1 occurrence
- C1 = Chorus 1 occurrence
- V2 = Verse 2 occurrence
- C2 = Chorus 2 occurrence
- V3 = Verse 3 occurrence
- C3 = Chorus 3 occurrence

The numbering represents the occurrence/order of the section in the song, not a different song-section type.

For example, C1 and C2 are both Chorus sections, but they are separate occurrences in the song's performance sequence.

---

## Example of the desired workflow

A worship leader or operator could create a song like:

V1
  Slide 1
  Slide 2

C1
  Slide 1
  Slide 2

V2
  Slide 1
  Slide 2

C2
  Slide 1
  Slide 2

V3
  Slide 1
  Slide 2

C3
  Slide 1
  Slide 2

ENDING
  Slide 1

The operator should have control over exactly where each slide changes.

---

## Why this would be useful

This would be particularly useful for live worship presentation.

Lyrics should not always be divided according to Verse/Chorus/Bridge alone.

A single Verse or Chorus can contain several lines, and the operator may want to divide those lines into smaller groups so that:

- The text is easier to read on a large screen.
- The congregation has enough time to read each group.
- The lyrics have better visual spacing.
- The presentation follows the actual musical phrasing.
- The worship operator can advance slides naturally during the song.
- Repeated Chorus sections can remain in their exact performance order.

---

## Important distinction

The proposed separator is a PRESENTATION SLIDE BREAK.

It should not change the underlying song section.

For example:

[Chorus]
DAYGON TA ANG DIYOS
MUSALIG KANIYA

[---]

ANG MALIPAYONG KASING KASING
HALLELUJAH

Should remain one Chorus section containing two presentation slides.

It should NOT be interpreted as:

Chorus
Bridge

or:

Chorus
Chorus 2

unless the user explicitly creates those sections.

---

## Suggested UI behavior

If possible, a visual "Add Slide Break" button in the Song Editor could also be useful.

For example:

[Add Slide Break]

This would insert:

[---]

The user could then visually control where the next presentation slide begins without having to manually type the syntax.

---

## Expected result

The user should have two independent levels of structure:

1. SONG SECTION
   - Verse
   - Chorus
   - Bridge
   - Ending
   - etc.

2. PRESENTATION SLIDE BREAK
   - Determines where the lyrics are divided on screen.

A slide break should not change the song section.

This would provide much better control for live worship presentations while keeping the song structure organized.

## Attach Screenshots 

Image

Image

zitlem:
Closes #404. **Stacked on #447** — that one is the base branch, so this diff is only the slide-break
work. Merge #447 first; GitHub will retarget this to `main`.

## The problem

A section that is too long for one slide had two options: leave it too long, or give the back half a
header of its own — which renames it, so half a chorus reads "Verse 2". The reporter's workaround,
a `[---]` line, was worse than inert: `ChordTransposer.isSectionHeader` accepts any bracketed line
that is not a chord, so the marker parsed as **a section named `---`**. Their screenshot on #403
shows exactly that — `---` badges, and a footer counting 7 sections.

## The change

`[---]` on its own line now ends the **slide** and not the section. Both halves keep the enclosing
`header`, `type` and colour, so they still read Chorus; `LyricSection` gains `slideIndex`/
`slideCount` so the operator is told which slide of the section they are on. The grammar lives in
`ChordTransposer.isSlideBreak` beside `isSectionHeader`, and is forgiving about how it is typed
(`[---]`, `{--}`, `[- - -]`) because this goes in by hand — while a section genuinely named
`[--- Chorus ---]` is still a section.

The consequences the plan flagged, each handled and each tested:

- **The chorus auto-repeat repeats whole sections**, so a two-slide chorus is sung as two slides
  every time round rather than losing its second half.
- **A break with nothing behind it costs no blank slide** — leading, doubled and trailing markers
  are what hand-editing actually leaves behind, and a blank slide is a black screen someone has to
  click past mid-service.
- **A break inside a chord-only intro folds away with it** rather than resurrecting it as an empty
  slide.
- **Slides are numbered after the chord-only fold**, so "2 of 2" counts what will reach the screen
  rather than what was typed.
- **Chords split with the words** — `chordLines` is divided at the same point, so the stage
  monitor's chart cannot drift from the lyric.
- **"Title on the first page" means the first slide**, not every slide of verse 1.

## Bilingual — the sharp edge, decided deliberately

`getLyricSections` paired the two languages on one running index, so a break in the primary and not
the secondary would have slid *every later section* under the wrong translation. Pairing now runs
per section and per slide within it: the section that disagrees comes out with an untranslated extra
slide, and verse 3 still meets verse 3. Matching breaks in both languages pair slide for slide. Both
are tested, and the existing index-pairing tests are untouched because an unsplit song is a run of
one-slide sections either way.

## Editor

A **Slide Break** chip beside the Intro/Verse/Chorus/Bridge chips, so nobody has to guess the
syntax — deliberately **not** added to `Constants.SONG_SECTION_MARKERS`, which is the list of things
that *are* sections. The marker is drawn as a divider while it is being typed, and the preview shows
the split with the same `CHORUS 1/2` badges the Songs tab uses. The footer still counts **sections**,
not slides: two slides of one chorus is one chorus.

## Tests

`:song-chords` — the marker is not a header, is recognised however it is typed, a dashed *name* is
not a break, and stripping leaves it alone. `:core-models` — it survives the file round trip,
including the end-trimming that would have eaten it. `:composeApp` — thirteen cases over the splitter
(two slides keeping one label, a three-way split, markers at the start/middle/end, a marker outside
any section, a song of nothing but markers, numbering across two split sections, the end-of-song
marker, the chord-only fold, the chart dividing with the words, a split chorus repeating whole, a
split verse collecting one chorus rather than one per slide), the two bilingual cases above, the
preview and its stats, the title-on-first-slide rule, and the editor chip driven through to the
saved song.

## Verification

`:song-chords:test`, `:core-models:test`, `:composeApp:check` and all three detekt gates are green,
as are the `:core-models`/`:song-chords` coverage gates (85% on all six counters, no new excludes,
no floors lowered).

Two new screenshot states show the feature — `songsTab/section_split_slides` and
`editSongDialog/slide_break` — and the `editSongDialog/*` set was re-recorded on macOS for the new
chip. `songsTab/` and `previewApp/` were otherwise byte-identical, since no existing fixture uses a
break.


zitlem:
Landed in #450, which was merged with this branch's commit `cf703abf0` in it — see `main` at 12a9d97dc. Closing rather than merging: after the stack was collapsed onto one merge, this branch's head is behind `main` and merging it would revert the per-section background work. #404 is fixed on `main`.

PR #456, opened 2026-08-31
Add NDI video input as a Canvas source

zitlem:
The Canvas can now receive an NDI source from the network as a layer, beside the camera and screen-capture sources — a camera on another machine, a graphics feed, an overflow room's output — over the same NDI runtime the NDI outputs already use.

## `:ndi` — the receive half

The module had deliberately left this out, and reserved this exact shape for it ("if it is ever wanted it belongs as a `SceneSource` in the Canvas compositor"). `NdiLibrary` gains discovery (find create/sources/destroy) and receiving (recv create/capture/destroy); `JnaNdiLibrary` binds them with three more ABI structs, and `NdiFinder`/`NdiReceiver`/`NdiSourceInfo` sit over them.

Three decisions worth reviewing:

- Receivers ask for `BGRX_BGRA`, so the runtime converts every YUV format before this code sees it. No pixel decoding in Kotlin.
- The frame copy honours `line_stride_in_bytes` — the runtime may pad rows, and assuming packed rows shears the picture. Packed frames still take a single read rather than 1,080.
- Every captured frame is freed, **including one that is dropped** for a format we do not read; it is the runtime's memory until it goes back. Audio and metadata are consumed and discarded: this is a video path, and a receiver that accepted audio nobody drained would grow the SDK's own queue for the length of a service.

## App side

- `NdiFrameCache` — one connection per distinct source however many layers draw it, reference counted like `SharedCameraFrameCache`, so the canvas preview and the presenter output do not make the sender encode the stream twice. Bandwidth is part of the key: two layers of one source at different bandwidths are two different streams.
- `NdiSourceDirectory` — one finder, held open while a picker is looking and closed afterwards. Discovery is cumulative in the SDK, so a finder created per query reports an empty network every time; one never closed keeps answering mDNS for a panel that shut an hour ago.
- The picker always offers the configured source even when discovery cannot see it — a camera switched off between services must not silently un-select itself from the scene the operator built.

## Two bugs the tests found while being written

- **Receivers were never closed.** The cleanup `withContext` ran inside the coroutine `release()` had just cancelled, so it threw instead of running: every NDI layer removed mid-service left the sender encoding for a subscriber that no longer existed.
- **Idle blanking was counted in polls**, which is two seconds against a live source and two milliseconds against one that answers instantly. It is measured against the clock now.

## Verification

`:composeApp:check`, `:ndi` and `:core-models` suites, coverage and detekt all green — `:ndi` keeps its 85% floor on all six counters with no new excludes.

Beyond the fake: the opt-in loopback test was run against the **real NDI SDK 6.3.2.0**. It put a sender up, found it through discovery as `AIS-MAC-MINI.LOCAL (ChurchPresenter Loopback Test)` at `127.0.0.1:5961`, connected, and read the frame back with its red channel intact. That is the only thing that can catch a wrong field order in the three new structs — a receiver built over a mislaid `NDIlib_source_t` connects to nothing and simply never gets a frame, with no crash and no exception.

## Screenshots

`add_source_menu` re-recorded for the new entry, and a new `source_ndi` state. Recorded on macOS, as the committed set is.

Two pre-existing failures are **not** touched by this branch: `StageMonitorScreenshotTest`'s clock churn (documented in AGENT.md) and `DictionaryTabScreenshotTest`, which fails only in a whole-suite verify and passes run alone, with and without these changes.


zitlem:
Merging this — the two findings below are follow-ups rather than blockers, but the first two are real native-resource bugs on ordinary UI interactions and are worth fixing soon.

**1. A cancelled connect leaks the native receiver.** `composables/NdiFrameCache.kt:140-142` — the open runs as `withContext(Dispatchers.IO) { openReceiver(...)?.takeIf { it.open() } }`. If the coroutine is cancelled while the blocking native `recv_create` is in flight, `withContext` discards the block's result and throws `CancellationException` before the value is ever assigned, so the `try/finally` that would close the receiver is never entered and the handle leaks. This is the same shape as the "receivers were never closed" bug the PR description says the tests caught, just on the connect side rather than the release side: add an NDI layer and remove it (or switch scene/tab) inside the connect window and the sender keeps encoding for a subscriber that no longer exists — once per removed layer, for the rest of the service.

**2. `NdiFinder` is used outside the lock that protects it.** `composables/NdiSourceDirectory.kt:62` — `sources()` reads `finder` under `synchronized(this)` but then calls `finder.sources(waitMs)` outside it, while `release()` (`:46-51`) calls `findDestroy` under the lock, and the field is not volatile. `NdiFinder` is documented as not thread-safe. So: the properties panel runs the 1000 ms `NDIlib_find_wait_for_sources` on IO, the operator clicks another layer within that second, the `DisposableEffect` disposes and calls `NDIlib_find_destroy` on a handle a native call is still blocked in. That is a use-after-free, and the window is a full second wide.

**3. `acquire` increments before it opens.** `NdiSourceDirectory.kt:38-42` — `refCount++` happens before `openFinder()`, so a failed open leaves `finder == null` with a non-zero count; every later `acquire` re-attempts `findCreate` while `release` only closes on the last one. Benign, but the ref count stops describing the finder's state.

**4. Per-frame allocation.** `NdiFrameCache.kt:168-190` allocates a fresh `BufferedImage` plus a fresh `toComposeImageBitmap()` per frame — roughly 8 MB x2 of garbage per 1080p frame, ~500 MB/s per layer at 30 fps. This mirrors the existing `SharedCameraFrameCache` shape so it is not a regression, but a reused buffer is worth it on a new path.

Minor: the new `FakeNdiLibrary` fields write `java.util.Collections.synchronizedList`, `java.util.concurrent.ConcurrentHashMap` and `ConcurrentLinkedQueue` fully qualified where imports would do (matches the file's existing lines, so: don't extend it); and `NdiHardwareTest`'s `HALF_RED` has a redundant double `.toInt()`.

What I checked and found sound: the three new ABI structs against `NDIlib_recv_create_v3_t` / `NDIlib_source_t` / `NDIlib_find_create_t` (JNA's 4-byte boolean is absorbed by pointer alignment on 64-bit, so the offsets line up — and the real-SDK loopback run in the description is the right way to have confirmed that); the stride-aware frame copy; `recv_free_video_v2` in a `finally` including the dropped-FourCC path; the `NonCancellable` close; the exhaustive `when` over `SceneSource` with all seven sites updated; and the bounded-poll `waitFor` helper, which is the wait style AGENT.md asks for.

PR #457, opened 2026-08-31
Read Keynote and legacy PowerPoint decks in the converter

zitlem:
The converter's **Documents** source read `.pdf`, `.docx` and `.pptx`. It now also reads Keynote `.key` and the legacy binary `.ppt`.

Worth noting up front: `.ppt` was never supported either. It reads as though it should have been — the source is described as "PowerPoint" throughout — but only OOXML `.pptx` was ever wired up, and `DocumentTextExtractorTest` quietly asserted the legacy *Word* sibling was out. Both legacy decks and Keynote are in now.

## Approach

The document path is already *extract text, then parse*, so each new format is one extraction function plus its entry in the two extension lists. Song splitting, section labelling and file writing are untouched.

Keynote goes through `:presentation-engine` rather than being parsed a second time. The IWA reader (snappy + a hand-rolled protobuf reader) was already there in full, but every type is `internal`, and the public `Deck`/`Slide` model carries `notes` and layer geometry with **no slide body text** — so the blocker was visibility, not capability. The new `KeynoteText.slideTexts` is the one text-shaped entry point:

- **Reading order.** Text boxes come back top-to-bottom then left-to-right. The parser emits z-order, which is creation order, so a title typed after the verse beneath it would otherwise sort second.
- **Groups** are flattened with the parent offset applied — children are stored relative to the group's origin.
- **Speaker notes are excluded.** They are the presenter's words; a song that folded them in gains verses nobody sang.
- **Fallback.** A deck that cannot be parsed natively — password-protected, or iWork '09 with an `index.apxl` body — falls back to the `QuickLook/Preview.pdf` Keynote embeds in every document it saves.

It keeps `PresentationLoader.load`'s contract: an unreadable deck is an empty list, never an exception.

## Package directories

A `.key` can be a package directory as well as a zip. `ObjectIndex.load` already accepted both, but the converter's folder scan filtered on `isFile`, so a bundle was skipped *and then walked into*. It is now taken whole as a single input, scoped to a `PACKAGE_EXTENSIONS` set so no other format's scan behaviour changes.

The file picker still enters a bundle rather than selecting it. Making bundles selectable needs a non-traversable `FileSystemView`, which would change the chooser for every format on that screen — so "select folder" is the remedy, documented at the site.

## Dependencies

`:converter` takes `projects.presentationEngine` and `poi-scratchpad` (HSLF is a separate hierarchy from the XSLF that reads `.pptx`). One-directional — `:composeApp` already depends on both. The single-POI-schema-jar rule is unaffected: the engine applies the same `poi-ooxml-lite` exclusion.

Only the English strings bundle is touched, and no keys are added, so no locale falls out of sync.

## Tests

Split by ownership. The engine owns the Keynote parsing cases, built with its existing programmatic IWA fixtures — both container forms, reading order, groups, notes, the preview fallback, and refusing to throw on damaged input. The converter owns dispatch and end-to-end, building a `.ppt` with HSLF and a `.key` whose preview PDF is written with PDFBox, so **no binary fixture is committed**.

`:converter:test`, `:presentation-engine:test`, both detekt gates, both coverage gates and `:composeApp:check` all pass. New suites were run three consecutive times with `--rerun-tasks`; no coverage exclude was added and no floor lowered.

## Not covered

- **No real Keynote file has been through this.** The fixtures are synthesized, so they prove the parser against the object graph the module already documents — not against what current Keynote actually writes. Worth one manual pass with a real deck, and one saved by iWork '09, before this is relied on.
- The engine's native path is exercised through synthesized IWA; the converter's `.key` tests reach it through the preview fallback only.

## Found, not fixed

`converter/build.gradle.kts` points detekt at `src/main/kotlin/converter`, but the sources moved to `src/main/kotlin/org/churchpresenter/converter/`. **That path matches nothing, so `:converter:detekt` currently analyzes only test sources** — its `AGENT.md` says "everything that parses a file is analyzed", and that has not been true since the move.

I verified this change is clean by temporarily correcting the path (zero findings in the files touched here). A real fix needs the `song`/`bible`/`library` subpaths rather than the parent, since the parent now also contains `ui/**`, which is deliberately out of scope — correcting it here would surface ~208 pre-existing `ui/**` findings and swamp the diff. Left for its own change.


zitlem:
Holding this one back from the batch — the package-directory support does not work, and I confirmed it by running it rather than by reading the diff.

## The `.key` folder scan yields nothing

`converter/src/main/kotlin/org/churchpresenter/converter/ui/App.kt:2701-2706`

```kotlin
dir.walkTopDown()
    .onEnter { it == dir || !isPackageInput(it, format) }
    .filter { it != dir && (it.isFile || isPackageInput(it, format)) && matchesFormat(it, format) }
```

`onEnter` returning `false` does not mean "yield it but do not descend" — it means the directory is **not visited at all**, itself included. So a `deck.key` bundle under the chosen folder is dropped entirely, and the `isPackageInput(it, format)` arm of the `filter` is unreachable.

Probe against the real stdlib, a temp dir holding `deck.key/Index/inner.iwa` and `plain.pptx`:

```
PROBE_RESULT=[plain.pptx]
```

`deck.key` is absent. So "select folder" over a folder of Keynote bundles returns zero inputs — the opposite of what the KDoc at the site and `converter/README.md` both promise.

`onEnter` cannot express the intent; this needs a hand-rolled walk, or `walkTopDown()` with the pruning done on the result sequence instead of at the enter hook.

Worth noting why CI is green anyway: `App.kt` is under `**/ui/**`, so it is outside `:converter:detekt` (after #458) *and* outside coverage, and `findFilesRecursive` has no test anywhere in `converter/src/test`. A test over a temp dir containing a bundle would have caught this.

## Smaller points

- **CI path filter.** `.github/workflows/test.yml:112-116` lists each of `:converter`'s project dependencies so the converter job reruns when they change. This PR adds `implementation(projects.presentationEngine)` but not `'presentation-engine/**'`, so a change to `KeynoteText` or the IWA parser will not run the converter suite that now depends on it.
- **Bogus size for a bundle.** `App.kt:698` renders `formatFileSize(it.length())`; `File.length()` on a directory is filesystem-defined, so a package would list as e.g. "96 B" once the walk is fixed.
- **Partial parse suppresses the fallback.** `KeynoteText.slideTexts` falls back to `QuickLook/Preview.pdf` only when *every* slide is blank (`native.any { it.isNotBlank() }`). A deck where slide 1 parses and the rest use an unhandled archive imports as one verse instead of falling back.
- **FQN in a new test.** `DocumentTextExtractorTest` uses `java.util.zip.ZipFile(...)` inline, against `CODING_STANDARDS.md` — and the same PR fixes exactly this for `SlideLayout` a few lines above.

## What is sound

The extraction itself. HSLF for legacy `.ppt`, `KeynoteText` going through `:presentation-engine` rather than parsing IWA a second time, reading order, group flattening, notes exclusion, and the no-committed-binary-fixture test approach are all good, and I verified the dependency direction is one-way.

I also checked this against #458 specifically, since the two touch the same module: the `converter/build.gradle.kts` hunks are disjoint and merge cleanly, and with both applied `:converter:detekt` analyses **64 files with 0 findings** (it was analysing zero before #458). So the interaction is fine — it is only the walk that needs fixing.

zitlem:
Pushed fixes for everything raised above, on top of a merge with current `main` (which now includes #458, so this code is detekt-gated for the first time). Four commits, kept separate so the walk fix reads on its own.

**The walk.** `onEnter` cannot express "yield this directory but do not descend into it", so the scan is hand-rolled now: a package is taken whole and never recursed into, a plain directory is recursed into, anything else is matched on extension.

**Moved it where the gates reach it.** This is the part that matters more than the fix. `App.kt` is under `ui/**` — outside `:converter:detekt` and outside the coverage floor — and `findFilesRecursive` had no test anywhere, which is the whole reason a feature returning zero results went green. The scan is now `converter/song/InputDiscovery.kt`: the file chooser stays in the UI, the decision about what counts as an input does not. `:converter:detekt` goes from 64 files to **66, 0 findings**.

The new suite is checked to actually bite — reverted to the original `onEnter` implementation, two of its cases fail:

```
InputDiscoveryTest > a keynote package directory is taken whole() FAILED
InputDiscoveryTest > packages and plain files are found together, at any depth() FAILED
```

It also pins the surrounding decisions: a bundle's innards never leak into the results, a directory merely *named* `archive.pptx` stays a folder to search, and a `.key` is only a package for a format that actually reads it.

**Bundle size**, which rode along in the same file: the selected-files card rendered `File.length()`, filesystem-defined on a directory, so a package would have listed as "96 B" beside its neighbours' real sizes once it started appearing at all.

**Partial parse.** Gaps are now filled per slide from the preview PDF, and only when the page and slide counts agree so an index cannot put slide 4's words on slide 2. Where they disagree the old all-or-nothing rule stands, and a deck that reads cleanly never opens the preview.

**CI path filter.** Added `'presentation-engine/**'` to the converter job, so a change to `KeynoteText` or the IWA parser runs the suite that now depends on it.

**FQN.** `ZipFile` imported rather than named inline.

Green locally: `:converter:` and `:presentation-engine:` test, detekt and `jacocoTestCoverageVerification` — no floor lowered and no exclude added — plus `compileKotlinJvm` and `:composeApp:detekt`. Worth noting `:presentation-engine:detekt` caught my first version of the fallback at 4 returns against a limit of 3; restructured rather than suppressed.

One thing deliberately left alone: the file picker still enters a bundle rather than selecting it. Making bundles selectable needs a non-traversable `FileSystemView` that would change the chooser for every format on that screen, and now that the folder scan actually works, "select folder" is a real remedy rather than a claimed one.

Scope note for review: only `DocumentFormat` carries `.key`, so of the three `findFormatInputs` call sites only the Documents folder scan changes behaviour — the other two formats walk exactly as before.

PR #458, opened 2026-08-31
Make the converter's detekt gate analyse its main sources again

zitlem:
`:converter:detekt` has been analysing **only its test sources**, and nothing said so.

`source` named `src/main/kotlin/converter`, carving `ui/**` out by simply not listing it. That was correct when the layout was `src/main/kotlin/{converter,ui}/`. When the sources moved under `org/churchpresenter/`, the path stopped matching anything — and a detekt run over zero files is a *passing* detekt run, so the gate went quiet instead of red. The module's own `AGENT.md` has been claiming "everything that parses a file is analyzed" throughout.

This is the module's first gate in CI (`.github/workflows/test.yml` runs detekt before any test), and it has no baseline, so it is meant to be the strictest one here.

## The fix

Name what is **excluded**, not what is included:

- `source` is now the whole of `src/main/kotlin` plus `src/test/kotlin`
- `**/ui/**` is excluded on the task

Two reasons this shape is better than repairing the old path:

1. **Every other module of this build already does it this way.** All 17 siblings use their whole source root; `:converter` was the only one narrowing, and so the only one that could break like this.
2. **The failure mode inverts.** A stale *include* silently analyses nothing. A stale *exclude* analyses too much and tells you. The gate can no longer switch itself off.

`ui/**` stays out for the reason it always did — pre-existing Compose desktop, the same shape `:composeApp` keeps in a baseline rather than gating, and this module has no baseline.

## One finding, fixed

Turning the gate back on surfaced exactly one: `VideoPsalmConverter.sectionsOf` had two `continue`s in one loop (`LoopWithTooManyJumpStatements`).

The repeat check keeps its short-circuit: `alreadySung` is still only touched for a verse that has lines, and `labelOf` still only claims a number for a section that is actually kept. Sections and labels are unchanged, and the VideoPsalm suite passes untouched.

## Verification

Beyond the suite passing, I checked the gate actually bites, since "detekt passes" was exactly the misleading signal before:

- injected a `MaxLineLength` violation into `song/` → **detekt fails** and names the file
- injected the same into `ui/` → **detekt still passes**, so the carve-out is intact
- reverted both

`:converter:detekt`, `:converter:test` and `:converter:jacocoTestCoverageVerification` all pass.

## Note for reviewers

This is independent of #457 and based on `main`, but the two touch the same module. Once both land, the converter code in #457 is analysed for the first time — I checked it against this corrected config while writing it, and it produces no findings.


zitlem:
Merging this. I verified the central claim rather than taking it on trust: on a tree with this branch applied, `:converter:detekt` reports **64 kt files analysed, 0 code smells**. Before it, the task was analysing zero files and passing. The inversion argument in the description is right — a stale include is silent, a stale exclude is loud.

One thing to fix in a follow-up: **the exclude also drops two test files from the gate.** Detekt's `exclude` patterns are matched against each source root, and `source` is now `["src/main/kotlin", "src/test/kotlin"]`, so `**/ui/**` matches `converter/src/test/kotlin/org/churchpresenter/converter/ui/SongSourcesTest.kt` and `.../ui/StringBundleTest.kt` as well. Those two were in scope before this PR and are not now, while `converter/AGENT.md` as shipped here still says `src/test/kotlin` is analysed. It is the same class of bug the PR is written to prevent, one level down — an exclude quietly matching more than intended. A main-only carve-out (a predicate on `/src/main/`, or scoping the exclusion to the main detekt task) would restore them.

I also checked the interaction with #457, since both touch `converter/build.gradle.kts`: the hunks are disjoint (dependencies block vs detekt block) and merge cleanly, and the `VideoPsalmConverter.sectionsOf` refactor is behaviour-preserving — `alreadySung` is still only touched for a verse with lines and `labelOf` still only claims a number for a section that is kept.

PR #459, opened 2026-08-31
Fix Keynote slides rendering the wrong number of text lines

zitlem:
Two decks reported by a user rendered with the wrong line count:

| Deck | Slide | Keynote | Before | After |
|---|---|---|---|---|
| Deck A | 1 | 4 lines | 5 | 4 |
| Deck B | 4 | 6 lines | 1 | 6 |

Both were downstream of parsing — the `.key` files decode to 4 and 6 paragraphs correctly — and a third defect kept either fix from reaching the screen.

## The extra line: PostScript font names never resolved

Keynote stores the typeface as a **PostScript name** (`Arial-BoldMT`, `HelveticaNeue-Bold`, `TimesNewRomanPS-BoldMT`), never as the AWT family name. `SlideFontRegistry.resolveFamily` only ever matched family names — and its substitution table is keyed on Office family names too — so a Keynote request missed both and fell through to the `Open Sans` default. That is not a cosmetic substitution: the fallback is a different width, so it silently re-wraps slides. Measured with the engine's own `LineBreakMeasurer` at the parameters `drawParagraphs` passes (38pt bold, 952.8pt wrap width):

| font | line 2 | lines |
|---|---|---|
| Arial (asked for) | 922.0 pt | 4 |
| Open Sans (actually used) | 987.0 pt | 5 |

This affected essentially every Keynote deck, not just this one.

`PostScriptName` peels a name back towards its family a step at a time (`Arial-BoldMT` → `Arial`, `HelveticaNeue-Bold` → `Helvetica Neue`, `TimesNewRomanPS-BoldMT` → `Times New Roman`), and each candidate is tried against the installed families *and* the substitution table, so a name still reaches its existing row on a machine that lacks the exact face. `resolveFace` additionally returns the weight and slant that only lived in the name; the rasterizer ORs those on top of `CHAR_PROPS_BOLD`/`ITALIC` rather than replacing them.

The family index is now also built lazily on resolve, exactly as `isFamilyAvailable` already did. Without that the engine resolved *everything* to `SansSerif` outside the app — the converter, the companion server's background renders, and the whole test suite.

## The missing lines: a character style leaking forward

Deck B's slide 4 is the only storage in either deck carrying a character-style attribute table, which is why it alone broke. Two defects in `KeynoteDeckParser`:

- **`attributeRuns` dropped runs with no style object.** In iWork such a run is not padding — it is the point at which the previous run's override stops applying. Dropping them let a character style covering the first word apply to all six paragraphs.
- **`resolveCharProps` returned a non-null, all-null result** for a style archive whose properties message is present but empty. That counted as "resolved" and suppressed the paragraph-style fallback wholesale, so paragraphs 2–6 landed on the hard defaults — 20pt, `Color.BLACK` — and rendered black on a black slide.

Character properties now merge over paragraph properties **per property**, so a character style that overrides only italic still takes its family, size and colour from the paragraph. Paragraph runs keep their carry-forward semantics (Keynote writes one per paragraph and names the style only where it changes) while character runs honour the clear.

## Why neither fix was visible: the slide cache never expired

`SlideDiskCache.lookup` validated an entry on manifest schema version, source mtime and render width only. An untouched deck therefore kept serving the images it was *first* rendered with, for ever — so both fixes above changed nothing on screen while the pre-fix JPEGs sat on disk.

`RENDERER_VERSION` describes the pixels as `schemaVersion` describes the manifest around them, and a mismatch is a cache miss. Entries written before the field existed decode as `0` and re-render once, so users need not clear anything by hand. **Bump it in the same commit as any change to what a rendered slide looks like** — layout, fonts, colour, fidelity gating.

## Verification

- Both decks rendered through `PresentationLoader` + `DeckRasterizer` before and after, and checked against the slide thumbnails Keynote itself wrote into the packages. The table at the top is that comparison.
- New parser tests build the real storage's table shape from `Fixtures`: an object-less character run that must clear rather than leak, and a character style that declares one property while the paragraph style supplies the rest.
- New font tests cover PostScript-name resolution, the style bits a name carries, and resolving before anything has initialised the registry. The exact-family assertions use the bundled Open Sans so they do not depend on the platform's font set.
- New cache tests: an entry behind on renderer version is invalid despite an untouched source, and a manifest with the field absent is invalid.
- `:presentation-engine:` test, detekt and `jacocoTestCoverageVerification` green; `:composeApp:` compiles with detekt clean.

## Not addressed

Real fidelity gaps in this path that did not cause either report, each worth its own change: no text insets on the wrap width, no shrink-to-fit, no vertical alignment, no paragraph spacing (`lineSpacing`/`spaceBefore`/`spaceAfter`), and one font per paragraph — so a mid-paragraph style change is lost at layout as well as at parse.


zitlem:
Merging this — the three fixes are real, correctly diagnosed and well tested. Two follow-ups on the name peeling, which now governs PPTX fallback as well as Keynote (`SlideFontRegistry.kt:227` routes every unavailable PPTX family through `resolveFamily` too), so the blast radius is wider than the Keynote-only framing suggests.

**1. Candidate order is not "most specific first", and the KDoc example is wrong.** Tracing `candidates("TimesNewRomanPS-BoldMT")` by hand: `splitCamelCase(base)` is `"Times New Roman PS"`, and `dropStyleWords` strips `PS` **and then `Roman`** — both are in `STYLE_WORDS` — giving `"Times New"`, which is emitted *before* `"Times New Roman"`. Inert in practice, since no machine has a family called "Times New" so it falls through to the right candidate, but it is luck rather than design. The documented sequence is wrong twice over as well: `Times New Roman Ps` is never produced (case is preserved as `PS`) and `Times` is never produced. Emitting all un-stripped forms before any stripped ones would fix both.

**2. `dropStyleWords` strips width and weight-class words.** `STYLE_WORDS` contains `narrow`, `condensed`, `expanded`, `light`, `black`, `book`, `medium`, `thin`, `heavy`, so `Arial Narrow` on a machine without it peels to `Arial` — a much wider face. Same for `Segoe UI Light`, `Helvetica Neue Thin`, `Franklin Gothic Book`. This only fires when the family is genuinely absent, where the pre-PR behaviour was an equally-wrong Open Sans, so it is not a regression — but re-wrapping is precisely what this PR exists to stop, and swapping one wrong width for another is not a fix. Consider excluding the width/weight words, or applying `dropStyleWords` only to forms derived from the hyphen suffix.

**3. A hyphen is only ever a style separator, never a space.** `candidates` seeds only from `substringBefore('-')`, so a family whose real name contains a hyphen loses everything after it. Adding `raw.replace('-', ' ')` to the seed list is a one-liner.

**4. The lazy index build has two edges.** It is data-race-safe (`ConcurrentHashMap` + `putIfAbsent`, idempotent refresh; worst case is duplicated work), but the lazy path never sets `initialized`, so it never scans system font directories and a later `initialize()` still does the full scan; and on a headless box where `getLocalGraphicsEnvironment()` throws, `availableFamilies` stays empty and every `resolveFace` in the render loop re-attempts and swallows the exception. An `AtomicBoolean` around the index build covers both.

**5. Ordering against rendering is still unguarded, and `RENDERER_VERSION` cannot rescue it.** `main.kt:311-320` runs `registerFontStream` + `initialize()` on a daemon thread, so a background render can start first; the lazy path only calls `refreshAvailableFamilies()`, which reads AWT families without registering system font files. A render that beats the scan resolves to a substitute, commits its JPEGs, and the entry then validates for ever — mtime, width and `rendererVersion` all match. This is the one failure the new cache field cannot detect.

Small ones: `KeynoteDeckParser.kt:433`'s `paraStyle?.let { resolveCharProps(index, paraStyleId) }` ignores `it` and `resolveCharProps` already null-checks, so the guard is redundant; `resolveCharProps` still sets `sawAny = true` for a present-but-empty properties message, which is harmless only because the merge is now per-property; `CharProps.isComplete` is dead code; and `attributeRuns` uses `lastOrNull { it.first <= charIndex }` — last in file order rather than greatest index — which now mis-*clears* rather than merely mis-picks if a table is ever out of order, so a `sortedBy { it.first }` is cheap insurance.

Two things worth stating in the PR text: the cache check is `!=`, so a version *downgrade* also re-renders; and `RENDERER_VERSION` is global, so this Keynote-only fix discards every cached PPTX and PDF deck as well. The "bump it in the same commit" rule should live in `presentation-engine/AGENT.md` rather than only in a KDoc, or the next renderer change will not know it exists.

Verified as sound: the manifest change is backward compatible (single reader with `ignoreUnknownKeys`, single writer, `cleanupOrphaned` keys off directory names, nothing outside the module decodes it, absent field decodes as `0` → one re-render); and the char/paragraph merge genuinely preserves carry-forward for paragraph runs while clearing for character runs — `paraStyleAt` is exactly equivalent to the old `styleAt` over the old filtered list, so the asymmetry is additive by construction. No conflict with #457: the presentation-engine file sets are disjoint and `KeynoteText`'s extraction reads only `paragraphs[].text`, which this PR does not touch.

PR #461, opened 2026-08-31
Camera backgrounds (#439)

### LINKED ISSUE #439: Backgrounds: allow a live camera as the background behind songs and verses
Split out of #384 (item 1). #385 was closed pointing here for this specific ask.

## What's missing

`BackgroundConfig.backgroundType` is one of Default / Color / Image / Video / Transparent / Gradient
(`settings/src/main/kotlin/org/churchpresenter/settings/utils/Constants.kt:59-65`). There is no
Camera.

A live camera can already be a **Canvas** source — `SceneSource.CameraSource`, covering USB capture
cards, webcams (DirectShow / V4L2 / AVFoundation) and DeckLink inputs with format and connection
selection. What it cannot be is the **background behind a song or a verse**, which is what was asked
for. Today the only way to get a camera behind lyrics is to build the whole slide as a Canvas scene,
which means giving up the Songs and Bible tabs entirely.

## Shape of the change

**Settings** (`:settings`)

- `Constants.BACKGROUND_CAMERA = "Camera"`.
- New defaulted fields on `BackgroundConfig`, mirroring the device identity `SceneSource.CameraSource`
  already carries: `cameraDevicePath`, `cameraDeviceName`, `cameraVideoFormat`,
  `cameraVideoConnection`, `cameraIsDeckLink`, `cameraDeckLinkIndex`. These are plain strings and
  ints, so `:settings` gains no dependency on `:composeApp`.
- No `SettingsManager` migration is needed — the format is kotlinx JSON with
  `ignoreUnknownKeys = true`, and defaulted fields are backward-compatible by construction.

**Rendering**

Draw it as an **overlay composable, not a `bgModifier` branch.** A camera is a live `StateFlow` frame
like video, not a static bitmap, so it wants the existing `useVideoBackground` shape —
`Modifier.background(Color.Black)` on the base plus an overlay drawn above it — rather than
`Modifier.paint(BitmapPainter(...))`. Touch points:

- `presenter/BiblePresenter.kt` (resolution ~403-445, `bgModifier` ~454-467)
- `presenter/SongPresenter.kt` (the same block, ~259-301 / ~310-323)
- `PresenterScreen.kt` (~38-73) for the default/screen-level slot

Note the resolution logic is currently **copy-pasted** between the two presenters, with a third
independent path in `PresenterScreen`. Extracting it into one shared function first would be a
reasonable prelude to this work, and would stop the camera branch being written three times.

**Frame source**

Acquire through the existing `composables/SharedCameraFrameCache`:

```kotlin
internal fun acquire(source: SceneSource.CameraSource): CameraFlows  // frame: StateFlow<ImageBitmap?>, error: StateFlow<CameraFailure?>
fun release(source: SceneSource.CameraSource)
```

Mirror `SceneSourceRenderer.kt`'s `CameraSourceContent` exactly: `remember` and `DisposableEffect`
keyed on the **same four device fields** (`devicePath`, `videoFormat`, `videoConnection`,
`deckLinkIndex`). This matters for two reasons:

- The cache is ref-counted and keyed only on device identity, so a background and a Canvas source on
  the same device **share one capture** — no second ffmpeg process, no second `openInput`.
- **The trap:** if the background stores `videoFormat = ""` while a Canvas source has
  `"1920x1080@30"`, the keys differ, you get two entries, and the second capture usually fails with
  `DEVICE_BUSY` after 5 retries. For DeckLink it is worse — two entries on one index fight over the
  hardware input. So the background must persist the device fields byte-identically to what the
  picker produced.
- `release` is keyed by the source you *pass*, not what you acquired with; changing format between
  acquire and release leaks a capture. Same-keys-both-places is the whole defence.

Pass `showDiagnostics = false` on the presenter path so camera failure text never reaches the
congregation. `ContentScale.Crop` matches how the other background types fill.

**Settings UI**

- Add a Camera entry to `BackgroundTypeDropdown` in
  `dialogs/tabs/BackgroundSettingsTab.kt`, and device / format / connection dropdowns to
  `BackgroundColumn` (line 471).
- Reuse the pickers from `composables/SceneDeviceEditors.kt` (`CameraProperties`) and
  `composables/DropdownSelector.kt` rather than writing new ones.
- Gate the entry on `isFfmpegAvailable()` the way `BACKGROUND_VIDEO` is gated on VLC, via the
  existing `disabledTypes` parameter.
- `listCameraDevicesWithDeckLink()` **blocks** (it shells out to ffmpeg / powershell /
  system_profiler). The Canvas UI calls it inside `remember` on the composition thread; don't repeat
  that here — do it on `Dispatchers.IO`.

## Open questions, to settle during implementation

- **Device missing or in use.** What does the output show — black, or the configured
  `backgroundColor`? A camera that is unplugged mid-service must not leave a frozen last frame.
- **Off-screen outputs.** Should a camera background be permitted on Browser Source and NDI outputs
  at all? Both go through `presenter/OffscreenOutputContent.kt`, which duplicates the mode dispatch
  rather than reusing `PresenterModeContent.kt` — so any threading has to land in **both** files.
- **Instance Link cannot mirror this.** `server/RemoteApply.kt` caches a mirrored background's image
  and video *files* locally; a device path is meaningless on the follower machine. The follower must
  fall back to its own background rather than render nothing.

## Not in scope

No `BACKGROUND_SLOT_*` work in `server/MediaAndAssetRoutes.kt` — that route serves background asset
*files*, and a camera has none.

zitlem:
Closes #439. **Stacked on #460** — review that one first; this branch contains its commit.

A camera could already be a Canvas source but not the background behind a song or a verse, so the only way to get one behind lyrics was to build the whole slide as a Canvas scene and give up the Songs and Bible tabs. `Camera` now sits alongside Color / Image / Video / Transparent / Gradient on every background surface, on a quick-tray tile, and on a song's own background.

## Approach

**One type carries the device identity.** `CameraDeviceRef` (in `:core-models`) with a single conversion to the `SceneSource.CameraSource` the shared frame cache is keyed on. The capture is ref-counted on device identity — path, format, connection, card index — and hardware opens *once*: a background storing `videoFormat = ""` beside a Canvas layer pinned to `1920x1080@30` would be two keys for one device, a second ffmpeg process, and `DEVICE_BUSY`. One conversion means the two doors cannot disagree, and `SongBackgroundCameraTest` asserts that against the real `keyFor` rather than a restatement of it. That is also why the format picker ships now rather than later.

**Black is the whole failure story.** Absent, already held, or unplugged mid-service all arrive as a null frame and all draw black. The cache clears its frame on release, so an unplugged camera cannot leave its last picture frozen on a projector. `CameraBackground` has no `showDiagnostics` parameter at all rather than one defaulted the wrong way.

**A stored path is not an identity.** A `.song` file and a settings export both travel, and `avfoundation://0` written elsewhere names *a* camera here — a different one, pointed somewhere else in the room. Opening it is worse than showing nothing, so `cameraResolves` matches on the device name (card index for a DeckLink). Instance Link drops a mirrored camera outright for the same reason.

**Enumeration never runs on the composition thread.** `listCameraDevicesWithDeckLink` shells out to ffmpeg / PowerShell / `system_profiler`; it goes through `CameraDeviceCatalog` on IO. Every tile and chip draws black rather than a live frame — a preview per tile would hold the device open for as long as the panel is on screen.

Opacity, dim and blur all apply. Browser Source and NDI get it for free: both delegate to `PresenterScreen` and the presenters, which is where the change lands.

## Also here

- `DeviceInfoReportTest` was scanning the whole diagnostic report for five-digit port strings, including the build's git short hash. `2adc18080` contains `18080`; it would have failed for anyone whose hash collided. Fixed at the source.

## Not in scope

No `BACKGROUND_SLOT_*` work — that route streams background asset *files*, and a camera has none. No live preview in the panel's 16:9 stage; noted as a follow-up.

## Verification

```
./gradlew :core-models:test :settings:test :song-chords:test :composeApp:check
./gradlew :composeApp:detekt :core-models:detekt :settings:detekt :song-chords:detekt
```
all green; new tests pass on three consecutive `--rerun-tasks` runs and the slowest is 0.36s. Screenshots re-recorded on macOS: the type segments gain a sixth (the row now wraps) and the song panel gains a fourth category pill, plus a new `type_camera`. The stage-monitor, Bible and Songs typography images also re-recorded and were reverted — none of them shows either control.

Not yet exercised against real hardware; the shared-capture and unplug checks in the plan need a machine with a webcam.


zitlem:
Merging this — the findings below are follow-ups, not blockers. Verified alongside #456 on a tree with all of them applied: `compileKotlinJvm`, `:composeApp:detekt` and `:composeApp:check` (plus `:core-models`, `:settings`, `:ndi`) are green together, and there are no conflicts with #456 in `strings.xml`.

**1. The new `PresenterScreen` branch has no `isSet` guard.** `PresenterScreen.kt:133` calls `CameraBackground(camera = card.camera, …)` directly, while `PresenterBackground.kt:113`'s `usesCamera` requires `type == BACKGROUND_CAMERA && camera.isSet` and `:278` gates on it. So a Camera-typed surface whose device has not been chosen yet acquires the shared cache under a blank key and draws permanent black on the idle output, instead of falling through to a colour. The two doors disagree, which is exactly the class of thing `CameraDeviceRef` was introduced to prevent.

**2. `withoutCameras` leaves the type set.** `server/RemoteApply.kt:546-553` clears the `CameraDeviceRef`s but leaves `defaultBackgroundType`/`backgroundType` as `"Camera"`, so the stated behaviour — "dropping it puts the follower on its own configured background" — does not hold. An Instance Link follower mirroring a primary on a camera background shows black for the whole service, indistinguishable from a dead output.

**3. `cameraResolves` is only applied on the song path.** `SongPresenter.kt:508` uses it; the settings-scope backgrounds (Default, Default lower-third, Bible, Song global) gate on `camera.isSet` alone. That leaves the exact scenario `CameraDeviceCatalog`'s own doc calls out: an imported settings file naming `avfoundation://0` opens *a* camera on the new machine, putting the wrong live picture behind the lyrics rather than showing black.

**4. Two tests reach outside the JVM.** `BackgroundSettingsTabScreenshotTest`'s camera state renders `CameraPickerRow`, whose `LaunchedEffect(Unit)` runs `CameraDeviceCatalog.refresh(...)` — shelling out to ffmpeg / PowerShell / `system_profiler` — and leaves the process-wide catalog at `[]`. `PresenterCameraBackgroundTest`'s last assertion depends on that same catalog still being `null`, so the pair is order-dependent within a fork. `BackgroundSettingsControls.kt:288`'s `isFfmpegAvailable()` effect adds the same cost to every existing screenshot of that tab. AGENT.md rules out both the shared-singleton ordering dependency and a unit test whose cost is an external process.

**5. A test name asserts the opposite of what it says.** `` `a song's camera is dropped when this machine does not have the device` `` asserts `assertTrue(resolved.usesCamera)`. The "dropped" behaviour has no presenter-level coverage — only the pure predicate in `CameraDeviceCatalogTest`.

Minor: `PresenterCameraBackgroundTest`'s `val BLACK = androidx.compose.ui.graphics.Color.Black` is a fully-qualified name where an import exists; `selectedBackgroundCameraName` (`BackgroundCameraPicker.kt:179`) calls `devices.first()` unguarded; `CameraBackground.kt:44`'s comment claims the remember keys are "the same four fields the cache hashes" but `SharedCameraFrameCache.keyFor` also branches on `isDeckLink`, which is not among them — unreachable today, but worth enforcing rather than asserting; and two stray blank lines at the end of `BackgroundSettingsControls.kt`.

The model side is sound and I checked it specifically: the six camera keys round-trip through `songBackgroundKeys`/`cameraFrom`/`cameraFields` including the `-1` DeckLink sentinel, fields are appended rather than reordered, blanked outputs drop the camera, ref-counting means N outputs on one camera open one capture, and enumeration stays off the composition thread.

Issue #497, opened 2026-09-08
Windows: EXCEPTION_ACCESS_VIOLATION in skiko swapBuffers freezes the window


zitlem:
Sentry: `CHURCH-PRESENTER-DESKTOP-52` — 4 events, 2 users, first seen 2026-08-29, still recurring. Status "regressed".

```
RenderException: Native exception in [Java_org_jetbrains_skiko_redrawer_WindowsOpenGLRedrawerKt_swapBuffers],
code 3221225477: EXCEPTION_ACCESS_VIOLATION
```

Windows amd64, on `AWT-EventQueue-0`, while presenting. The stack is entirely skiko/AWT — no application frames — so it is a GPU driver access violation surfaced by skiko as a Java exception.

Not fixed: it cannot be reproduced or verified without an affected machine, and `SOFTWARE` rendering would be a serious regression for a dual-output presenter app. Recording what the investigation found so whoever picks it up does not start cold.

### We chose OpenGL on Windows ourselves

`composeApp/build.gradle.kts:629` sets `-Dskiko.renderApi=OPENGL` for the packaged Windows build, overriding skiko's own Direct3D default. That is why this lands in `WindowsOpenGLRedrawer` at all. **Direct3D looks like the first fallback rung to try, well before software rendering.**

Note skiko's `fallbackRenderApiQueue` only covers *context creation* failure. A `swapBuffers` fault happens long after that, so skiko's built-in fallback never fires here.

If a runtime override is wanted, `main.kt:237-240` is the place — it already does exactly this for macOS/Metal, and `System.setProperty` beats the packaged `-D` flag because skiko reads the property lazily on first `SkiaLayer` creation.

### Moved out, and done

Three sections that lived here have been resolved or split off — see the comment below for detail:

- **It is a freeze, not an exit** → split to #518. The uncaught handler never calls `exitProcess`,
  so a fatal exception on the AWT thread leaves a live process with a window that has stopped
  repainting. Applies to any such exception, not just this fault, and is reachable without an
  affected machine.
- **It currently blames the wrong component** → fixed in #511. The handler now records a
  `CrashKind` classified by exception type, and a renderer fault no longer trips
  `videoBackgroundsDisabled`.
- **The "re-enable" affordance is dead** → fixed in #511. `composables/CrashGuardBanner.kt` has
  real buttons, and `reEnableVideoBackgrounds()` has a production caller.

[...]

PR #511, opened 2026-09-09
Stop forcing OpenGL on Windows, and stop blaming VLC for GPU faults

zitlem:
Refs #497. **Deliberately not `Fixes`** — see "What this does not do" below.

## Why

Sentry `CHURCH-PRESENTER-DESKTOP-52`: `RenderException: Native exception in
[Java_..._WindowsOpenGLRedrawerKt_swapBuffers], code 3221225477:
EXCEPTION_ACCESS_VIOLATION`, on `AWT-EventQueue-0` while presenting. 4 events, 2 users,
status regressed. The stack is entirely skiko and AWT — no application frame in it.

## The four changes

**1 — Windows no longer pinned to OpenGL.** `build.gradle.kts` set
`-Dskiko.renderApi=OPENGL` in both the dev `run` task and the packaged `windows { }` block,
overriding skiko's own Direct3D default. That override is the only reason the fault landed in
`WindowsOpenGLRedrawer.swapBuffers` at all. skiko's `fallbackRenderApiQueue` covers *context
creation* only, so a fault during `swapBuffers` never triggers it — the default has to be
right rather than recoverable.

Linux keeps its pin: it guards against software fallback and is unrelated to this.
`shouldForceMetalRenderer` becomes `preferredRenderApi(osName, override)`, with
`CHURCHPRESENTER_RENDER_API` as the operator's way out if a machine fares worse on Direct3D
(documented in `README.md`, mirroring the existing `CHURCHPRESENTER_FORCE_DEV_WINDOW`).

**2 — Reports now say which renderer was live.** A crash report carried no renderer, GPU or
adapter information, so change 1 was unfalsifiable. New `render.api` tag, set right after
`CrashReporter.initialize` (before the availability tags — a render fault can arrive first),
plus the same line in `DeviceInfoReport`.

**3 — A GPU fault no longer disables video backgrounds.** `evaluateCrashEscalation` took only
`(crashedLastRun, previousCount)`. Any run that failed to delete `.running` counted, and two
in a row set `videoBackgroundsDisabled`, gating the whole VLC decode path. A GPU driver fault,
a force-quit, an OOM kill and a power loss were indistinguishable from a VLC crash.

The handler now records a `CrashKind`, classified by exception **type** per
`diagnostics/AGENT.md` ("classify by exception type, never by message"), and a renderer fault
no longer trips the guard. **A crash that records nothing still does** — that is precisely
what a native VLC fault looks like, and it is the case the guard was built for. The kind is
consumed and cleared each startup, so a stale one never judges a later crash.

**4 — The crash banner works.** It had never been visible: emitted before `MainDesktop`,
whose opaque `fillMaxSize` root painted over it. `[Re-enable]` and `[Dismiss]` were literal
characters inside one `Text` with no button behind either — which is why
`reEnableVideoBackgrounds()` had **zero production callers** — and its only interaction was an
`onPreviewKeyEvent` on a non-focusable `Text`, so nothing but the 15s auto-hide ever fired.
Now `composables/CrashGuardBanner.kt`: a top-aligned strip with real buttons, drawn after the
content it sits over, using string resources.

## What this does not do

**It does not prove the access violation is fixed.** That needs an affected machine, and none
is available — Please leave #497 open.


zitlem:
Reviewed and tested on real Windows hardware — Windows 11 Pro, JDK 21.0.10 Temurin, GeForce GTX
1660 Ti (driver 31.0.15.5186), 32 GB. Four commits pushed to this branch.

## The branch could not start on Windows

`./gradlew :composeApp:run` died before the first window:

```
java.lang.Exception: Windows does not support Metal rendering API.
	at org.jetbrains.skiko.SkikoProperties.parseRenderApi$skiko(SkikoProperties.kt:146)
```

`jvmArgs` is declared **only** on `JvmApplication` — no platform settings class has it (checked
against the plugin bytecode). So `macOS { jvmArgs("-Dskiko.renderApi=METAL") }` never applied to
macOS alone: it resolved against the enclosing `application { }` receiver and went to every
platform, the Windows `run` task included. Windows escaped only because the `when` above pinned
OPENGL *first* and Gradle keeps the first value for a repeated `-D` key. Removing that pin — the
point of this PR — left METAL as the only value.

Confirmed by running both sides and reading the live process command line:

| branch | `-Dskiko.renderApi` on the JVM command line | result |
|---|---|---|
| `main` | `OPENGL` (METAL deduped away) | window opens |
| this branch, before the fix | `METAL` — the only value left | crash at startup |

**Fixed in cf2c0d0** by making `preferredRenderApi` the single source of truth for all platforms and
setting no render API in `build.gradle.kts` at all. That also removes a quieter problem: the `when`
read the os.name of the machine doing the *build* and baked it into the artifact. Decided at
startup, the answer is right whoever built it and wherever it runs. Only `-Xdock:name` stays in the
build file, guarded on a Mac host, since a launcher argument can't be set from inside the JVM.

Verified from the loaded driver DLLs:

| run | flag on cmdline | `D3D12Core.dll` | `nvoglv64.dll` |
|---|---|---|---|
| default | *(none)* | **yes** | no |
| `CHURCHPRESENTER_RENDER_API=OPENGL` | *(none)* | no | **yes** |

Also fixed: a blank `CHURCHPRESENTER_RENDER_API` masked a deliberately passed
`-Dchurchpresenter.renderApi`, because `System.getenv` returns `""` for a variable set to nothing.
Windows deletes such a variable outright; macOS and Linux do not, which is where it bites.

## Two production bugs found while getting the suite green

- **c5dd04c** — VLC reports its own default output with an empty id and an untranslated description
  ("Default"). Since the stored `audioOutputDeviceId` is also `""`, it *matched*, so the closed
  button showed VLC's English string instead of the localized `audio_output_default`, and the
  dropdown carried two rows meaning the same thing. In a non-English locale the control read
  "Default" in English.
- **c3d0825** — `MediaViewModel.deriveTitleFromUrl` fell back to `substringAfterLast("/")` for a
  local file that no longer exists. A Windows path has no forward slash, so the **entire**
  `C:\Media\clip.mp4` became the media title.

## Suite: 211 failures → 0

`:composeApp:check` now passes on Windows — **11062 tests, 0 failures**, detekt clean.

The handoff notes predicted the 169 `Projection*` failures were a missing sound card and would pass
on a box with audio. That is inverted. This machine has 7 audio devices and VLC installed, and they
still failed — `awaitAudioDevices` waited for text the button could never show while VLC's empty-id
entry was matching the empty stored id. They pass on a machine with **no** audio and fail on one
with it. The same root cause as c5dd04c above.

The rest (a0d405b) were tests asserting things true only on POSIX: raw Windows paths interpolated
into JSON literals (invalid escapes → unparseable document → the `400` and the null cache),
`/chosen/...` POSIX literals round-tripping through `Path`, `detectVlcInstallPath` asserting against
the developer's own disk, `ffmpeg` vs `ffmpeg.exe`, and the `STTTabTest` race. Each fixed at what
made it machine-specific rather than by loosening the assertion.

So `DEVELOPMENT_GUIDE.md`'s claim that Windows runs the full suite **is now true**; it was not before.

## Still open — filed rather than left in a comment

- #515 — the crash banner has no screenshot and no test yet, per `AGENT.md`'s "approve the UI
  first" rule. It needs a design decision: it's a `TopCenter` overlay, so it sits **over** the
  toolbar rather than pushing content down.
- #514 — Direct3D is confirmed to initialise and run, but **not** under the lyrics-plus-looping-video
  dual-output load the original crash happened under. The "Direct3D is better here" claim is
  untested under real load, on NVIDIA only.
- #513 — a follower re-separates the primary's remote path with its own separator (display-only).
  Pre-existing, unrelated to this PR; the assertions here were adjusted to tolerate it.

Keeping `Refs #497` rather than `Fixes` is right — #514 or a production report carrying the new
`render.api` tag is what can settle it.

Issue #507, opened 2026-09-08
Media Features Request


tobsowo:
For the media feature, can you please add
1. Option to loop video: We play a video on a loop before the service begins
2. Subtitles: The ability to add subtitles to a video and the possibility of making them multilingual
3. Media Uploads: Upload and display images and videos on different screens per language

grunt2001c:
You should be able to play video loops already using backgrounds, if background is not a thing using media, or Canvas. With canvas you can add any text on top of the video.

tobsowo:
Thank you for your response.
Adding a loop/repeat button will actually make the UX better.

<img width="299" height="77" alt="Image" src="https://github.com/user-attachments/assets/8241f44b-c5e5-426e-8f66-c241ab1a0ac9" />

zitlem:
Splitting this so the small one can move independently:

- **Item 1, loop a video** → #523, being worked on now. You are right that a loop button on the
  transport row is a better fit than sending the file through backgrounds or the Canvas: those loop
  today, but you give up the Media tab's own controls to get it.
- **Item 2, subtitles** — stays here. Substantially bigger: a subtitle track has to be selected,
  rendered on the audience output rather than only in the preview, and the multilingual part means
  more than one track live at once.
- **Item 3, per-language media on different screens** — this is the same underlying capability as
  #475 (a language per output). Better tracked there than solved twice; the app is bilingual today
  (primary/secondary) and routing *n* languages to *n* outputs is one piece of work, not one per
  content type.

Keeping this issue open for item 2.

tobsowo:
Thank you for adding the subtitle.
Is it possible to use the subtitle track to display on different screens?
Is it possible to change the subtitle font, colour, size, outline, etc.?

zitlem:
Opened #592: subtitles are now app-drawn and fully styleable (font, colour, size, position, bold/italic/underline, shadow, backdrop, outline), from a new gear icon next to Go Live on the Media tab. Embedded/.ass/.ssa/.sub tracks stay on VLC's own rendering, unstyled — the libVLC API that would let the app style those too only exists from a VLC 4 requirement, which isn't reasonable to impose app-wide for this.

tobsowo:
I guess the gear icon did not meet up with the nightly build

<img width="1172" height="340" alt="Image" src="https://github.com/user-attachments/assets/943df259-a4b6-48d5-ae12-3dcdc32d95a7" />

PR #527, opened 2026-09-10
Let the Media tab repeat what it is playing

### LINKED ISSUE #523: Media tab: a loop button on the transport controls
Split out of #507 (item 1), which also asks for subtitles and per-language media routing — both much
larger, and the third overlaps #475. This is the small, separable one.

## The ask

> Option to loop video: We play a video on a loop before the service begins

The reporter's follow-up, after being pointed at video backgrounds and the Canvas:

> Adding a loop/repeat button will actually make the UX better.

with a screenshot of the Media tab's transport row as the place it belongs.

## Where it stands today

The Media tab has no loop concept at all — `MediaViewModel` has no such state, and a file that ends
simply stops. Looping *does* exist elsewhere in the app, which is what makes this a gap rather than
a new capability:

- Video **backgrounds** loop (`composables/LoopingVideoBackground.kt`, via `SceneVideoSpec(loop = true)`).
- **Presentations** have `isLooping`, persisted in `PresentationSettings`.

So the workaround the reporter was offered is real — put the video in as a background, or on the
Canvas — but it means giving up the Media tab's transport controls to get a repeat, which is the UX
point being made.

## Scope

- A loop toggle in the Media tab transport row, matching the existing toggle-style controls there.
- The setting persisted, as `PresentationSettings.isLooping` already is.
- Looping honoured on the audience output, not only in the tab's own preview.
- Audio as well as video — the same transport drives both, and a looping bed before a service is the
  same request.

## Not in scope

Subtitles and per-language media routing (#507 items 2 and 3, the latter overlapping #475).

zitlem:
Fixes #523, split out of #507.

> We play a video on a loop before the service begins … adding a loop/repeat button will actually
> make the UX better.

A loop button on the Media tab's transport row, in the same place and the same shape as the ones the
Pictures and Presentation tabs already have — the same `ic_refresh` icon, the same tinted-when-on
treatment, the same `loop_on`/`loop_off` strings. **No new strings and no new icon asset**, so no
locale file is touched.

## The one design decision worth reviewing

The established way to loop video in this repo is VLC's own `:input-repeat=65535`, passed as a media
option at `play()` — that is how `LoopingVideoBackground` and canvas video layers do it, and it is
seamless because libvlc handles the repeat internally.

**I did not use it here**, because it is fixed for the life of the playback. The option is applied
when the media loads, so an operator who turns looping *off* half way through a video would watch it
repeat anyway, with no way to stop it short of hitting stop. For a control the operator can toggle
while something is live, that is the wrong failure.

Instead the decision is read at the end of the file: `markFinished()` checks the toggle, and if
looping, rewinds and bumps a `replayVersion` that both players key their load effect on. Turning it
on or off part-way through takes effect on the video that is playing.

Reloading rather than seeking is deliberate — at the end of its media the player has **stopped**,
and seeking a stopped player leaves it stopped.

**The trade-off:** libvlc's internal repeat is seamless, and a reload may show a brief gap at the
loop point. Verified working in the running app before this was committed. If the gap turns out to
matter for a pre-service loop, the upgrade is to layer `:input-repeat` on for the common case (loop
already on when the file loads) and keep this path as the fallback for a mid-play toggle — but that
is two mechanisms, and I would rather add the second one only if the first proves not good enough.

## Two things that follow from how media ends

- `mediaFinished` is what `main.kt` turns into `requestClearDisplay()`. A looping file must not
  raise it, or the audience output would black between repeats.
- Playback stays marked as playing, so the transport does not flip to a paused-looking state each
  time round.

Both are asserted.

## Where the setting lives

`AppSettings.mediaIsLooping`, beside `mediaStorageDirectory` — there is no `MediaSettings` class
(the settings dialog's "Media" tab edits picture and presentation settings, despite the name), and
one boolean does not earn a new group. A new field with a default needs no migration step.

**It defaults off**, unlike `PictureSettings.isLooping` and `PresentationSettings.isLooping` which
default on. A slideshow reaching its end and starting again is the ordinary expectation; a sermon
clip or a testimony video quietly restarting is not.

## Verification

- 9 new tests: 5 on the view model (the blanking flag, the replay counter, the toggle read at end of
  file in both directions, the default) and 4 on the tab (initial state, toggling, persistence,
  seeding from settings, and that it works before anything is loaded — it is not gated on `isLoaded`,
  because choosing to loop and *then* picking the file is the order setup happens in).
- The Media view-model, tab and presenter suites run three consecutive times with `--rerun-tasks`,
  all green.
- `:settings:test` green; `:composeApp:detekt` and `:settings:detekt` clean.
- Screenshots re-recorded: 7 in `mediaTab/` and 2 in `previewApp/`.

**Worth knowing about the screenshot gate:** `verifyRoborazziJvm` **passed** on the unrecorded
images even though 7 of 8 genuinely changed. A 28dp icon button plus its divider falls under the
0.1% `CHANGE_THRESHOLD`. So the gate does not catch a small control being added — the images here
were re-recorded because I checked, not because verify asked.

## Harness change

`MediaReports` gained `settingsAfterChange` (the shape `PresentationReports` already had) and the
harness now seeds its `MediaViewModel` from the settings under test, which a toggle reading its
initial value from settings needs.


zitlem:
Closing in favour of #525, which does the same feature and does it better — it was open before this one and I did not check before starting. Duplicated effort, my mistake.

Two things #525 handles that this PR gets wrong or lacks:

- **The Stage Monitor's own decoder.** It mounts a second `SoftwareVideoPlayer` on the same view model, so `markFinished()` can fire more than once per play. #525 stops that at both ends — `reportsPlaybackEnd = false` for a mirror, and a `playbackGeneration` guard against a doubled event. This PR would have reloaded once per reporting decoder.
- **Loop count**, and a dedicated `looping.png` screenshot for the on-state.

The one thing here that #525 does not have is **persistence** — its `isLooping` starts false every launch, and #523 asked for the choice to be saved the way `PresentationSettings.isLooping` is. Pushing that onto `video-loops` as a follow-up commit rather than leaving it here.

Branch kept for now in case anything else in it is wanted.

PR #564, opened 2026-09-17
Fix Lottie lower third floating centered on portrait outputs

### LINKED ISSUE #519: More than two text labels in a lower third
**Is your feature request related to a problem? Please describe.**
More than two text labels in a lower third (today it is exactly two).

**Describe the solution you'd like**
There are suitation we would want to display the song title, composer, writer and singer

zitlem:
## Summary
- On a portrait output, `ContentScale.Fit` against a landscape Lottie composition shrinks the whole graphic width-constrained, and it was left vertically centered (empty space above and below) because the outer `Box`'s `BottomCenter` alignment has no effect once `Image()` already fills that box with `Modifier.fillMaxSize()`.
- Sets `alignment = Alignment.BottomCenter` on `Image()` itself in `LowerThirdPresenter.kt` (all three render paths: live-painter, pre-rendered frame, and the frame-before-composition fallback) and in `LowerThirdOffscreenRenderer.kt` (ATEM media-pool pre-render), so the shrunk composition sits flush at the bottom instead of floating in the middle.
- 16:9 outputs are pixel-unchanged: the fitted image already fills the full canvas height there, so bottom vs. center anchoring makes no visible difference.

Fixes #519.

## Test plan
- [x] `./gradlew compileKotlinJvm`
- [x] `./gradlew :composeApp:jvmTest --tests '*LowerThirdPresenter*' --tests '*LowerThirdOffscreenRenderer*'` — all pass unchanged
- [x] `./gradlew :composeApp:detekt` — clean
- [ ] Manual: set a portrait output resolution, go live with a Lottie lower third, confirm it's now anchored to the bottom of the output (live window, and NDI/Browser Source if feasible) and that a normal 16:9 output is unchanged


zitlem:
Also fixed a second bug found while manually testing this: a screen assignment slot pointing at a specific display index kept its stale bounds forever once that display was unplugged, which silently overrode the dev fallback window's resolution picker (`outputSizeOf` prefers non-zero `targetBoundsW/H` over `devWindowWidth/Height`). Now reconciled the same way an auto (-1) slot already was, plus a regression test for it.

